PR #20, opened 2017-02-19
initial import

guillomovitch:
Here is a new shiny Nessus analyser for Cortex.

Here is the list of configuration parameters:

- url: the nessus scanner URL
- login: the nessus scanner user name
- password: the nessus scanner password
- ca_bundle: an optional CA certs file for validating nessus scanner certificate
- allowed_networks: an optional list of networks (CIDR format), to restrict the scanner usage to those networks only

### DIFF
--- a/analyzers/Nessus/Nessus.json
+++ b/analyzers/Nessus/Nessus.json
@@ -1,0 +1,11 @@
+{
+    "name": "Nessus",
+    "version": "1.0",
+    "baseConfig": "Nessus",
+    "config": {
+        "check_tlp": false
+    },
+    "description": "Nessus scanner",
+    "dataTypeList": ["ip", "fqdn"],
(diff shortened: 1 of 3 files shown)


tob1no:
Hey there,
I've got a question about the ca_bundle option.
I have Nessus Professional running with a locally issued certificate. How can I add the CA in Cortex?
I'm running the Nessus analyzer in thehiveproject/cortex:3.1.7 and mounted the CA bundle using the following volume switch:
 - "./data/ca-bundle.crt:/etc/pki/tls/certs/ca-bundle.crt"
 Now when I specify /etc/pki/tls/certs/ca-bundle.crt as the ca bundle path in the Nessus analyzer config, it still gives me certificate errors trying to contact Nessus Pro. What am I missing?
 
 Thank you very much in advance!

Issue #4, opened 2017-02-16
Cut python 2 dependency by replacing ioc-parser in cortexutils.analyzer


3c7:
### Request Type
Feature Request

### Content
Hello guys,

I just skimmed through analyzer.py to see, if it's possible to use the `Analyzer`-class with python 3. The problematic dependency in this case is `pdfminer` which is not in use, basically, because the parser is always called with `txt` and `json` parameters.

Would if be possible to replace it with an self-implemented function (more or less) easily?

Lots of modules I depend on use python 3 and I just thought it would be great to directly start with python 3 here.

Have a nice day
3c7

nadouani:
Closing this issue, fix by [Automatic ioc extraction using RegEx](https://github.com/CERT-BDF/Cortex-Analyzers/pull/40)

PR #40, opened 2017-04-18
Automatic ioc extraction using RegEx

3c7:
This is a basic extractor for ioc's. It searches through the `raw` dict and try to match some RegEx rules. This is no fulltext search and is only capable of finding something like `[{'results': '127.0.0.1'}]` but not `[{'results': 'The result is 127.0.0.1'}]`. Unittests included 🍻.


3c7:
bildschirmfoto vom 2017-04-18 08-46-14

3c7:
Also interesting for #4.

PR #47, opened 2017-05-10
Unit777/analyzers

bullerdude:
New analysers for Cortex:, initial versions:
- WhoisXMLAPI (Tool: https://www.whoisxmlapi.com/whois-api-doc.php)
- IRMA (Tool: http://irma.quarkslab.com/


nadouani:
Hi guys, I've just tested the IRMA analyzer with an IRMA 1.5.2 VM:
- Since Cortex 1.1, analyzers require 3 more information: `author` (author name or org name), `license` and `url`.
- I don't have any experience with IRMA, and I'm wordering if one could enable an authentication mechanism to protect the IRMA APIs, and in this case the analyzer should allow the users to provide at least creads for basic authentication (that was my case, since I've got access to the demo instance demo quarkslab)

Please let know if you want me to add the basic authentication support, otherwise the analyzer is OK, we just need to review the `summary()` method to produce a taxonomy for mini reports

BrevilleBro:
Hi @nadouani, we can update the IRMA analyser when we get some time, unless someone else is happy to take over development of it as it is no longer in use for us.

nadouani:
@BrevilleBro  that's fine, I'll add what I was talking about. I've reviewed the analyzer and it sounds fine. We're just missing the `author` information ;)

3c7:
Hey @bullerdude @BrevilleBro,
how is the status on this one?

BrevilleBro:
Hi @3c7 ,

We are no longer using these and therefore have ceased development of them. They should however, be compatible with the current versions of Cortex, as @nadouani has kindly cleaned them up.

PR #49, opened 2017-05-15
Update misp_analyzer.py to optionally work with self signed misp certificate.

MariasStory:
Added option to cortex conf for misp, to ignore ssl check for self signed misp certificate.
Add /etc/cortex/application.conf in MISP object:
ssl = "False"

Any value that is not True will disable ssl check.


jeromeleonard:
This option is included in the future version of the analyzer submitted by @3c7. The option is named `certpath`.

PR #50, opened 2017-05-19
Cuckoo Sandbox Analyzer

garanews:
Analyzer that will submit files to a local Cuckoo Sandbox instance and retrieve the report that is generated.


saadkadhi:
Hi @garanews 

Thank you for your submission. Can you please add an `author` (Real name and/or Org name), `license` and `url` to your JSON files? These are now required by Cortex 1.1+. 

Please note that if the value of `license` is not `AGPL-v3`, you'll have to add a LICENSE file to the analyzer folder that contains the full text of the license you have selected to release your analyzer under.

Examples:
[https://github.com/CERT-BDF/Cortex-Analyzers/blob/master/analyzers/DomainTools/DomainTools_ReverseIP.json](https://github.com/CERT-BDF/Cortex-Analyzers/blob/master/analyzers/DomainTools/DomainTools_ReverseIP.json)
[https://github.com/CERT-BDF/Cortex-Analyzers/blob/master/analyzers/CIRCLPassiveDNS/CIRCLPassiveDNS.json](https://github.com/CERT-BDF/Cortex-Analyzers/blob/master/analyzers/CIRCLPassiveDNS/CIRCLPassiveDNS.json)

Thanks in advance.

saadkadhi:
@garanews We either need `author`, `org name` or both. So if you can provide an `author` but not an `org name`, that's totally fine.

garanews:
Hi, now it should be ok.

PR #67, opened 2017-06-27
Fixed mistake in blocklist script, added error on missing config

3c7:
There was a pretty obvious copy/paste mistake in the code. I don't really know, how that mistake made it through the testing. Merged an older commit or so?! Anyway, the fix + removed default value for path config. Think it's clearer if there's an error message telling the user no path was set instead of using one that is probably not there.


Popsiclestick:
+1

PR #72, opened 2017-07-10
Builtins Added

peasead:
The `builtins` Python module is required and it is included in the `future` Python package.

I added `future` to the `requirements.txt` files.


3c7:
Why is the builtins python module required for every Analyzer?

nadouani:
`future` is already included in cortexutils that is required by all analyzers, and not all analyzers use `future` and `builtins`.

@peasead let me know if you agree with that, and we will decide for the PR.

PR #73, opened 2017-07-11
Added VxStream Sandbox (Hybrid Analysis) Analyzer

yugoslavskiy:
According to data from official site [1], VxStream Sandbox Public API
allows you to analyze:

- hash (OK, it works)
- filename (OK, it works)
- host / ip (some problems on API side for now)
- domain / fqdn (some problems on API side for now)

[1] https://www.hybrid-analysis.com/apikeys/info


saadkadhi:
@yugoslavskiy We'll rename the analyzer to HybridAnalysis instead of VxStreamSandbox to align with the fact that it indeed interact with the public service and that is called `Hybrid-Analysis_GetReport` in the service interaction file (.json).

yugoslavskiy:
@saadkadhi np. 
Also you probably will be interested in their own [cortex analyzer](https://github.com/PayloadSecurity/VxCortex) (released 8 days ago)

PR #80, opened 2017-07-19
Updates to Virusshare analyzer

colinvanniekerk:
I've added case insensitivity to the hash comparison within the analyzer. This is important because Virusshare stores all lowercase hashes, but if you compare an uppercase hash, it will always returns as not within the Virusshare dataset which can be dangerously incorrect.

I have benchmarked the comparison and found that the removal of the strip function actually improved the overall speed of the test when a line contains a single hash. Moving to the 'in' operator also allows for support of the UPX hash map file that has been created.

---

I've also added a simple bash script to fetch the hashes since I had issues with python version that is available. Bash is simpler and more or less always works as expected whereas python could apparently have issue that are harder to fix than (in this case) simply writing a bash based tool.

### DIFF
--- a/analyzers/Virusshare/getHashes.sh
+++ b/analyzers/Virusshare/getHashes.sh
@@ -1,0 +1,29 @@
+#!/bin/bash
+# This script downloads all available Virusshare.com hash files using curl and wget. It can be called as: ./getHashes.sh PATH
+
+
+display_usage() { 
+    echo "getHashes v0.1" 
+    echo "  Fetch all Virusshare.com hashes" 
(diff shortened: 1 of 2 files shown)


3c7:
I think my python script that downloads the hashes can be deleted. That bash script is much more cleaner.

3c7:
Think you have to modify the templates as well, because you changed `unknown` to `Unknown`.

colinvanniekerk:
You are correct :) I missed that, sorry.

I have implemented a better (lazier) fix so that the templates are not
affected... I generally prefer not to have case in internal marker strings
anyway. :P

PR #86, opened 2017-08-03
Adding some analyzers for common tools

alexgoedeke:
This pull requests has the following analyzers:
- Bluecoat: Check domains against Bluecoat's site reputation check
- MailParser: Parser *.msg files and check IPs. URLs and attachments against MISP, also run the FileInfo analyzer against every attachment
- ThreatCrowd: Run some basic checks against ThreatCrowd
- ipvoid: Search ipvoid for information.
- Th0r: Parse th0r reports and lookup outputs in MISP


co-devs:
Any chance for review?  An IPvoid analyzer would be a nice addition.

nadouani:
Hello, I'm personally not a fan of calling a Cortex service and a MISP from the analyzer code.

nadouani:
And since the PR is old, it depends on a Cortex v1 which has no authentication mechanism etc..

jeromeleonard:
I agree. for IPvoid there is the #455 that could be in a future release.

PR #142, opened 2017-11-21
Splunk Analyzer

BrevilleBro:


3c7:
Hey @BrevilleBro, thank you very much for providing the Splunk analyzer.

Unfortunately, it does not work with Python 3 - also owed by the splunk module itself which is not Python 3 compatible. We'd like to have all Analyzers working with Python 3, so maybe we can find a solution using some community forks of the original splunk-sdk package.

3c7:
There is an issue at the original splunk-sdk-python repository regarding Python 3 support where users also linked their forks which are ported to py3. We could give https://github.com/tltx/splunk-sdk-python/ a try.

BrevilleBro:
Hi @3c7, I am hesitant to use community forks, however, if Python3 compatibility is a must it may be necessary to do so.

3c7:
Either that or waiting for #171. Generally we would like to use Python 3 for all Analyzers and work towards that with the ones that are already released.

Best would be if the splunk team decides to use some compatibility package like six to support Python 2 and 3 - but that won't happen, I guess.

axpatito:
Hi @BrevilleBro as per https://github.com/splunk/splunk-sdk-python/commit/689998070b513a335efee1e4cc27fbf568b256f3 the offical library now has python 3.5 support. I've tested it with your code and it works.

The only other change required to support python3 is to call urllib's quote_plus method from urllib.parse.quote_plus.

I can PR if you want me to.

LetMeR00t:
Hi everyone, hi @3c7 ,
Do you still need help on this ? I can provide you the fixes of the concerning code here if you want as I'm using both Cortex and Splunk.

jeromeleonard:
finally decided to merge #534 which is an updated version of this one.

Issue #149, opened 2017-11-29
VirusShare downloader bash script bug


thebitbucket123:
### Title
Bash script to download VirusShare hashes wgets strings unrelated to hash files

### Request Type
Bug

### Work Environment
|---------------------------|--------------------
| OS version (server)       |  Ubuntu
| OS version (client)       | 16.04
| Cortex Analyzer Name      | VirusShare getHashes.sh
| Cortex Analyzer Version   | 0.1
| Cortex Version            | 1.1.4
| Browser type & version    | N/A

### Description
When pulling down virusshare's hashes using the bash script, the script parses lines past the end of the list of hashfiles. Currently, "https://virusshare.com/hashes.4n6" has a line at the bottom containing a link to "https://twitter.com/CodexGigasSys", which is parsed and wgat. No harm done here, but in the unlikely event that virusshare is ever popped and the page is changed to include a less-than-savory URL, we'd end up downloading that.

### Steps to Reproduce
1. Use the gethashes.sh script to download the VS hashes
2. Wait until the end

### Possible Solutions
 - Use a more specific regex, like /VirusShare_[0-9]{5}\.md5/, to parse or check to make sure we only get what we're looking for 

### Complementary information
Screenshot of the script attempting to wget the twitter URL
![virusshare_gethashes_bug](https://user-images.githubusercontent.com/25698662/33402004-df2abd7c-d520-11e7-864c-fccecf0617cd.PNG)

PR #154, opened 2017-12-13
Fixes bug in virusshare hash downloads

3c7:
Related to #149.


3c7:
Messed smth up, created a new one.

PR #171, opened 2018-01-17
WIP/evaluation: Using optionally Docker to prevent compatibility problems with Analyzers

3c7:
This is WIP and needs discussion.


BrevilleBro:
Good work @3c7 , I think this is the way forward for Cortex analyzers if we want to maintain compatibility.

ryanmcvicar:
Would this allow cortex analyzers to start running on different Docker images (Windows)? For example I could see a cortex analyzer that check Office 365 block lists which are only available over PowerShell at the moment.

geekscrapy:
@ryan I guess if you used docker swarm that'd be fairly easy to distribute analyzers across platforms

3c7:
- [ ] Fix cortex-base-python3 image
- [ ] Fix `unterminated quoted string` bug
- [ ] Check for possibilities to reduce image sizes

3c7:
Basically finished, just postponed to the next cortex (bugfix) release.

3c7:
Continued in #334 as I cannot change the branch.

PR #186, opened 2018-02-15
Bluecoat analyzer handling of multiple categories and their maliciousness level

koppensb:
Since Bluecoat can give URLs multiple categories and this analyzer just handled one I changed the way it works with them. 

The maliciousness level is set to the highest of the categories. In the summary funkction I set the level as I tought it would be most usable. Maybe the community thinks a little bit different, but we will see.

One problem i didn't fix is that the categories have IDs and in the report templates of thehive they are used for linking to the description of Bluecoat. For now only the first found ID is returned.

### DIFF
--- a/analyzers/Bluecoat/categorization.py
+++ b/analyzers/Bluecoat/categorization.py
@@ -26,8 +26,8 @@ class BluecoatAnalyzer(Analyzer):
         if categorization != "":
             result = {}
             try:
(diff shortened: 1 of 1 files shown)


3c7:
As the bluecoat analyzer was removed due to changes at bluecoat site and violation of their terms, I'm going to close this.

Issue #198, opened 2018-02-22
IPv4 address Extractor regex does not match only IPv4 address


srilumpa:
### Request Type
Bug

### Work Environment

| Question              | Answer
|---------------------------|--------------------
| OS version (server)       | Debian
| Cortex Analyzer Name      | cortexutils
| Cortex Analyzer Version   | 1.2.0
| Cortex Version            | 1.0.0, hash of the commit

### Description
The regular expression used to detect if an artifact is an IPv4 address is "too wide" and will detect as an IP address every string containing four groups of one, two or three digits separated by dots.

### Steps to Reproduce
1. Call `Extractor().check_string("1.0.0.127.localhost.localdomain.")`
2. This will return `ip` but should return `fqdn` (or an empty string)

### Possible Solutions
Use the following regular expression for ip type detection/extraction: `^\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3}(?:\/\d\d?)?$`

3c7:
Some thoughts in the PR (#199).

3c7:
I'll close this for now. The automatic extraction should be the one to use, if you've no idea of how to extract your observables properly and the behaviour can always be changed through overwriting the `artifacts` function. Maybe we need to add this to the "How to create an analyzer" manual.

PR #199, opened 2018-02-22
Cortexutils extractor ip detection

srilumpa:
Fix proposal for #198.

* 1.0.0.127.localhost.localdomain. => does not match
* 192.168.0.10 => does match
* 192.168.0.1/24 => does match
* 10.0.0.0/8 => does match
* 10.0.0.0/ => does not match

### DIFF
--- a/contrib/cortexutils/extractor.py
+++ b/contrib/cortexutils/extractor.py
@@ -28,7 +28,7 @@ class Extractor:
         # IPv4
         regex = [{
             'type': 'ip',
-            'regex': re.compile(r'^\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3}$')
+            'regex': re.compile(r'^\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3}(?:\/\d\d?)?$')
         }]
 
         # IPv6
--- a/contrib/cortexutils/extractor.py
+++ b/contrib/cortexutils/extractor.py
@@ -28,7 +28,7 @@ class Extractor:
         # IPv4
         regex = [{
             'type': 'ip',
-            'regex': re.compile(r'[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}')
+            'regex': re.compile(r'^\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3}$')
         }]
 
         # IPv6


3c7:
As the extractor should extract observables that aren't available as "a single line" also, line start and end markers (`^` and `$`) are not really applicable here and with that something like `1.0.0.127.localhost.localdomain` will be extracted as `1.0.0.127`.

I need to think about how (if?) it's possible to distinguish between an IP in CIDR notation which is in-line and the mentioned "domain case".

Rolinh:
Even with the fix, `999.888.777.666` would match as a valid IPv4 address when clearly, it is not.

Why do you have to use regular expressions to match IP addresses? Why not use something like [ip_address](https://docs.python.org/3/library/ipaddress.html#ipaddress.ip_address) from the standard library?

```python
def is_ip(s):
    try:
        ip_address(s)
        return True
    except ValueError:
        return False
```

In the end, there is not different types for IPv4 and IPv6 in TheHive, only `ip`.

Also, maybe it would be worth considering adding a `network` type to TheHive instead of having a range in CIDR notation considered as a valid IPv4 (and you guessed it, [ip_network](https://docs.python.org/3/library/ipaddress.html#ipaddress.ip_network) could be used to validate a network in CIDR notation).

3c7:
> Why do you have to use regular expressions to match IP addresses?

Because the extractor should provide an easy way to retrieve observables from reports even if they are in-line and not explicit given. Analyzers do not have to use the extractor and can implement an own `artifacts` method.


> Why not use something like ip_address from the standard library?

That would indeed be possible - after finding a possible IP address using regex.

Rolinh:
>>  Why do you have to use regular expressions to match IP addresses?

> Because the extractor should provide an easy way to retrieve observables from reports even if they are in-line and not explicit given. Analyzers do not have to use the extractor and can implement an own artifacts method.

I'm sorry, I don't understand why this makes using regular expressions a hard requirement.

>> Why not use something like ip_address from the standard library?

> That would indeed be possible - after finding a possible IP address using regex.

Again, why? `ip_address` can take a string as input so you could feed it with the value to check just as you feed `regexp.match`. Of course, this would imply a minor refactoring(*) of the `Extractor` class but in the end you would obtain more reliable results.

(*) The changes would be minimal and only internal to the class.

3c7:
> Again, why? ip_address can take a string as input so you could feed it with the value to check just as you feed regexp.match.

Yes, it takes a string, but cannot find addresses in a block of text. Again, the extractor should provide the functionality to "automagically" find observables/IoCs in strings which are **not** the observable itself, but a "wall of text".

Of course is recognizing single ip strings through regex is not the best way to achieve that. But you cannot guarantee that.

As this affects only analyzers run through MISP, it has a low prio for me until cortex 2 is released and the documentation is polished up as you're able to easily delete inappropriate IoCs (what has to be done anyway, as not every returned value is appropriate) in the overview after running the analyzer.

kx499:
I'm running into an issue where the IP's extracted (ipv4) include version strings and are not valid. I fixed this in code that calls cortex (external App) and basically filters the artifacts through the a call to IP Address. You could post process after the regex results are returned with ipaddress(val).is_global to limit out some of the noise.

PR #227, opened 2018-04-12
Add URLhaus analyzer

ninoseki:


3c7:
Hey @ninoseki, thanks for your contribution. 

It would be great to implement caching, so urlhaus/abuse.ch is queried only once even if analyzing lots of observables in the same time. For an example how to implement caching, you can take a look at the Tor analyzers available.

If you're feeling not familiar with implementing caching, I will implement it if I find some time.

ninoseki:
@3c7 how is this?

3c7:
@ninoseki Damn, I'm sorry. I thought you're working on the csv export data (their API, basically) from URLHaus. That's why I wanted to have some kind of caching mechanism added.

saadkadhi:
This analyzer, while very good, requires `requests-html` which is only compatible with Python 3.6. However Ubuntu 16.04 LTS and other OSes do not support yet that version and we cannot force everyone to upgrade to 3.6. 

We will have sadly to postpone it to another release after modifying the code to use the CSV export from URLhaus or have some kind of `requests-html` equivalent that works with Python 3.5. @ninoseki Are you willing to consider doing so?

Thanks!

ninoseki:
@saadkadhi Oh, it’s absolutely my fault.  Let me re-implement the analyzer without `requests-html`.

3c7:
Hey @ninoseki please see details written in #271.

Issue #231, opened 2018-04-17
Grey Noise analyzer


markus-nclose:
# GreyNoise analyzer

### Request Type
Analyzer

### Work Environment
N/A

### Description
[GreyNoise ](https://greynoise.io/)is a database that contains a list of IPs that is not malicious or potentially malicious. It is able to quickly tell you if an observable is safe or not. 

A quote from their site:

At GreyNoise, we collect and analyze untargeted, widespread, and opportunistic scan and attack activity that reaches every server directly connected to the Internet. Mass scanners (such as Shodan and Censys), search engines, bots, worms, and crawlers generate logs and events omnidirectionally on every IP address in the IPv4 space. GreyNoise gives you the ability to filter this useless noise out.

We are currently developing this analyzer.

PR #236, opened 2018-04-21
GreyNoise analyzer

danielbrowne:
Creating a new analyzer for GreyNoise.
https://github.com/GreyNoise-Intelligence/api.greynoise.io

Referencing #231

### DIFF
--- a/analyzers/GreyNoise/GreyNoise.json
+++ b/analyzers/GreyNoise/GreyNoise.json
@@ -1,0 +1,25 @@
+{
+    "name": "GreyNoise",
+    "version": "2.2",
+    "author": "Nclose",
+    "url": "https://github.com/TheHive-Project/Cortex-Analyzers",
+    "license": "APLv2",
+    "description": "Determine whether IP has known scanning activity",
+    "dataTypeList": ["ip"],
+    "baseConfig": "GreyNoise",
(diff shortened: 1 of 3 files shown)


3c7:
Hey @danielbrowne, thanks for your contribution! Do you have - by chance - already created a template for TheHive?

danielbrowne:
Hi @3c7, I have not done that. I am not that familiar with the product, is that a requirement for this work to be useful?

3c7:
As the main use case for analyzers is the query out of TheHive, it would be cool if you could provide either a template for it or an example of the data returned by the analyzer so we can create a template. Examples for templates can be found in https://github.com/TheHive-Project/Cortex-Analyzers/tree/master/thehive-templates.

If no template is provided, the users will just see a JSON object which is not that UX friendly. ;)

3c7:
Ah, just saw that the returned data is quite nice described in the the link you provided. So, do you want to try to create a template or should we do it?

danielbrowne:
@3c7 I am not sure that I will be able to get to it soon due to being flooded by other commitments, if it is easy for someone else to create it, that would be great :)

3c7:
Added a template. :)

danielbrowne:
Thank you :)

PR #239, opened 2018-04-27
JoeSandbox analyzers: use a sane analysis timeout

ant1:


ant1:
ping @saadkadhi

ant1:
ping @saadkadhi 
ping @nadouani

PR #271, opened 2018-06-06
Add URLHaus analyzer

3c7:
This is the urlhaus analyzer provided by @ninoseki (#227). It needs some refactoring, because the dependency `requests-html` seems to need Python 3.6 while Ubuntu 16.04 LTS comes with 3.5. As a large user base uses Ubuntu 16.04, this analyzer should be refactored.

Possible measures:
- use another module to parse the HTML site
- use the provided CSV export from URLHaus (**favoured**)

If you want to participate, @ninoseki, you can base PRs on the feature/urlhaus-analyzer branch, as your PR cannot be reopened, because it was merged before.

### DIFF
--- a/thehive-templates/URLhaus_1_0/long.html
+++ b/thehive-templates/URLhaus_1_0/long.html
@@ -1,6 +1,6 @@
 <div class="panel panel-info" ng-if="success">
   <div class="panel-heading">
-    URLhaus search resutls for
+    URLhaus search results for
     <strong>{{artifact.data}}</strong>
   </div>
   <div class="panel-body">


ninoseki:
I prefer scraping to the CSV export because URLhaus's CSV export doesn't contain hash values.
So I think scraping is better than the CSV export. 

@3c7 Does it make sense? 
If so I'll implement `beautifulsoup` ver. of this analyzer.

3c7:
Hey @ninoseki, there are two csv dumps, one containing the urls and one containing palyoad hashes:
- https://urlhaus.abuse.ch/downloads/csv/
- https://urlhaus.abuse.ch/downloads/payloads/

So it would be possible to use both and combine them in the result.

3c7:
Thanks for your contribution @ninoseki. :) I've made some changes to the report to display colored tags instead of text:

screenshot from 2018-07-02 10-08-03

PR #285, opened 2018-06-09
Feature/urlhaus analyzer

ninoseki:
@3c7 I’m so sorry for missing that API.(#271)
But I’m still sticking to the scraping because of these reasons:
- We could combine the 2 CSV exports results but it’s a little bit complex.
- When dealing with CSV,  we should implement matching code to multi-fields. It’s also a little bit complex.


ninoseki:
@3c7 @jeromeleonard Thanks!

Issue #116, opened 2017-10-22
Manalyze analyzer


lctrcl:
### Request Type
Analyzer

### Work Environment
N/A

### Description
Create analyzer for [Manalyze](https://github.com/JusticeRage/Manalyze), static analyzer for PE executables.

I created initial analyzer for Manalyze, using this quite a lot for quick malware triage. Feel free to try it out, it has very basic report template because my angular and html skills are not good. Requires docker. 

[https://github.com/lctrcl/cortex-analyzers](https://github.com/lctrcl/cortex-analyzers)

saadkadhi:
@lctrcl wouldn't it be possible to remove the docker requirement?

lctrcl:
@saadkadhi I certainly can, and it would require compile Manalyze from sources. But I also would like to keep docker option, as it's quite easy to deploy without managing additional requirements (except for having docker).
Should I create this as separate analyzers?

saadkadhi:
Thanks for your reply. Stay put. We'll look at the code and let you know if we can stay like this or not? At any rate, thanks for your contribution!

JusticeRage:
Hi! I'm Manalyze's original developer. Is there something I can do to help this happen? I would love to see the tool available in The Hive.

saadkadhi:
Ping @3c7

3c7:
The analyzer provided by @lctrcl looks fine. I'd prefer to let the user choose if using docker or the native binary. If using docker, I need to check if it works with the containerized analyzers in #171. I think it might be possible to implement a Manalyzer analyzer for Cortex-Analyzers 1.12.0.

3c7:
Currently in the process of integrating manalyze as FileInfo submodule.

3c7:
Finished and tested in #333 and ready for review/merge

PR #333, opened 2018-08-17
Manalyze submodule for FileInfo analyzer

3c7:
This adds a submodule to FileInfo analyzer. Either the manalyze binary or docker can be used and have to configured explicitly before using the analyzer.


3c7:
Ref: #116 
/cc: @JusticeRage, @lctrcl

3c7:
screenshot from 2018-08-17 12-46-10

- Coloring based on given level in Manalyze
- Docker and binary support

PR #351, opened 2018-10-05
Malware clustering

garanews:
This analyzer is used to correlate malicious samples that will be stored in Neo4j Database.
It supports Malpedia repository to have a known malware families baseline (more than 1000 families).
It uses ApiScout for correlation and g6.js for generate the chart.

image

thehive

Enjoy!

More info @ https://www.hack.lu/misp-summit/


3c7:
Thank you @garanews. Finally merging this - sorry for taking so long.

PR #358, opened 2018-10-18
Fix for Fortiguard to handle FQDNs as well as domains and urls

phpsystems:
The fortiguard analzer didn't accept fqdns. This fix changes that.


3c7:
Thank you @phpsystems!

PR #368, opened 2018-11-02
Additional features for IBM X-force plug-in

gekkeharry13:
Hi guys, as we were struggling during testing this plug-in, I added certificate verification options. And to actually use the plug-in we require proxy support. So hereby a pull request for adding both these features.

If there is anything not compliant to any standards, let me know so I can adjust it.

### DIFF
--- a/analyzers/IBMXForce/IBMXForce_Lookup.json
+++ b/analyzers/IBMXForce/IBMXForce_Lookup.json
@@ -37,6 +37,13 @@
       "required": true,
       "multi": false,
       "type": "string"
+    },
+    {
+      "name": "verify",
+      "description": "Enable/Disable certificate verification",
+      "required": false,
+      "multi": false,
+      "type": "string"
     }
   ]
 }
--- a/analyzers/IBMXForce/ibmxforce_lookup.py
+++ b/analyzers/IBMXForce/ibmxforce_lookup.py
@@ -2,6 +2,7 @@
 # encoding: utf-8
 import requests
 from datetime import datetime
+from urllib3.exceptions import InsecureRequestWarning
 
 from cortexutils.analyzer import Analyzer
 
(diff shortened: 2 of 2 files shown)


analyzers/IBMXForce/IBMXForce_Lookup.json
```diff
@@ -37,6 +37,13 @@
       "required": true,
       "multi": false,
       "type": "string"
+    },
+    {
+      "name": "verify",
+      "description": "Enable/Disable certificate verification",
+      "required": false,
+      "multi": false,
+      "type": "string"
```
nadouani: This can be ob type `boolean`, instead of making it a `string` and convert it to `boolean`

nadouani: Could be 
```
{
      "name": "verify",
      "description": "Enable/Disable certificate verification",
      "type": "boolean",
      "required": false,
      "multi": false,
      "default": true
    }
```

nadouani: I'll fix this change

gekkeharry13: Hi Nabil, sorry I thought you already adjusted the suggestion. I did not know there was a boolean option (could not find it for some reason, probably my fault) If this is changed, then we can remove the convert to boolean part as well

```
    def str2bool(self, v):
         if isinstance(v, str) or isinstance(v, int) or isinstance(v, unicode) :
             return v.lower() in ("yes", "true", "t", "1")
         if isinstance(v, bool):
             return v
```
If you are busy, I can try to fix it today? 😄

nadouani: @gekkeharry13 oh don't worry it was not a big deal ;) thanks for the contribution

Issue #382, opened 2018-11-30
Cisco Umbrella Blacklister Responder


arnydo:
### Request Type
Responder

### Description
Responder will allow the blacklisting of a domain in Cisco Umbrella utilizing the [Enforcement API](https://docs.umbrella.com/enforcement-api/reference).

### Possible Solutions
I'm working on the creation of the responder.

arnydo:
I am attempting to have the responder available for domain observables. I assume the configuration is the same as an analyzer. Is there something I am missing? The responder is not available when using TheHive.

`"dataType": "domain",`

![image](https://user-images.githubusercontent.com/11653079/49297173-251a2480-f488-11e8-8dee-03c452ebc3ce.png)

nadouani:
Hello, Responders are not like analyzers and can run against cases, tasks, observables, logs and alerts.

You can find the list at: https://github.com/TheHive-Project/CortexDocs/blob/master/api/how-to-create-a-responder.md#datatypelist

arnydo:
@nadouani Thank you for clarifying. I missed that page of the docs...

I have it working now.

Ideally, I would like to have the option to add a tag to the artifact showing that it has been blacklisted in Umbrella. Is there any ETA on adding the operation "AddTagToArtifact"? Should I go ahead and create a PR without it or wait until this is available?

[UmbrellaBlacklister](https://github.com/arnydo/Cortex-Analyzers/tree/responder/UmbrellaBlacklister/responders/UmbrellaBlacklister)

arnydo:
Please disregard last comment. Found that "AddTagToArtifact" is available; it just was not listed in the documentation. I created a PR for that change in CortexDocs.

Created PR for this: #383

nadouani:
PR accepted

PR #383, opened 2018-11-30
Responder/umbrella blacklister

arnydo:
Responder will allow the blacklisting of a domain in Cisco Umbrella utilizing the [Enforcement API](https://docs.umbrella.com/enforcement-api/reference).

Adds the tag "Umbrella:blocked to the observable.

Reference to issue #382


nadouani:
Hello @arnydo Thanks for the PR

PR #386, opened 2018-12-05
Adding Patrowl analyzer

MaKyOtOx:
Hi all,
Here is a proposal for an analyzer dealing with a Patrowl platform (see https://patrowl.io). Don't hesitate to reach me if you have any question or suggestion.

-- Nicolas

### DIFF
--- a/analyzers/Patrowl/Patrowl_GetReport.json
+++ b/analyzers/Patrowl/Patrowl_GetReport.json
@@ -1,0 +1,26 @@
+{
+    "name": "Patrowl_GetReport",
+    "version": "1.0",
+    "author": "Nicolas Mattiocco",
+    "url": "https://github.com/TheHive-Project/Cortex-Analyzers",
+    "license": "AGPL-V3",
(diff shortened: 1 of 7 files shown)


analyzers/Patrowl/Patrowl_GetReport.json
```diff
@@ -0,0 +1,26 @@
+{
+    "name": "Patrowl_GetReport",
+    "version": "1.0",
+    "author": "Nicolas Mattiocco",
+    "url": "https://github.com/TheHive-Project/Cortex-Analyzers",
+    "license": "AGPL-V3",
+    "description": "Get the current Patrowl report for a fdqn, a domain or an IP address.",
+    "dataTypeList": ["fqdn", "domain", "ip"],
+    "baseConfig": "Patrowl",
+    "config": {
```
nadouani: These conf properties should be made customizable, and declared as ConfigurationItems. Note that the anlyzer should use an API_KEY instead of username/password couple

PR #393, opened 2018-12-11
Improvement: Eml_Parser Analyzer & Template

arnydo:
Made small changes to how headers are parsed, storing in a dict rather than list.
Updated long template to better display the headers.


arnydo:
Finding some reports are not wrapping the text in the html table when the content is longer. Is there a pre-configured style than can be used to correct this? I was able to have it wrap the text by adding the following to the table element.

`style="table-layout: fixed;word-wrap: break-word;"`

`<table class="table table-striped table-bordered" style="table-layout: fixed;word-wrap: break-word;">`

Without additional style:
image

With additional style:
image

nadouani:
Reviewed this PR, thanks @arnydo

Issue #410, opened 2019-01-22
Crtsh Analyzer: crt.sh result is a nested list


sprungknoedl:
### Request Type
Bug

### Work Environment
N/A

### Description
The crt.sh result gets wrapped in a list when parsing the json, but the json returned from crt.sh is already a list. This results in a nested list:

Example (truncated for brevity):
```json
{
  "summary": {
    "taxonomies": [
      {
        "predicate": "Certificates",
        "namespace": "crt.sh",
        "value": "1",
        "level": "info"
      }
    ]
  },
  "full": {
    "certobj": {
      "domain": "example.com",
      "result": [
        [
          {
            "not_after": "2020-12-02T12:00:00",
            "min_entry_timestamp": "2018-11-29T13:44:14.118",
            "min_cert_id": 987119772,
            "issuer_ca_id": 1191,
            "name_value": "www.example.com",
            "issuer_name": "C=US, O=DigiCert Inc, CN=DigiCert SHA2 Secure Server CA",
            "not_before": "2018-11-28T00:00:00"
          }
        ]
      ]
    }
  },
  "success": true,
  "artifacts": [],
  "operations": []
}
```

nadouani:
This should be fixed by #411

PR #411, opened 2019-01-22
Fix for #410 removed wrapping of crt.sh result in a list

sprungknoedl:


3c7:
Thanks for your PR. I think there are some template changes necessary as the result value changes?

sprungknoedl:
The current result templates are not working (this is how I detected this issue), as they expect only a 1-level deep list

So the current templates match the result value of this PR

3c7:
Oh snap.

nadouani:
Approved by @3c7, merging the PR

PR #418, opened 2019-02-05
Updating Cuckoo Analyzer/Report Templates

nicpenning:
### Request Type
Analyzer

### Work Environment
N/A

| Question              | Answer
|---------------------------|--------------------
| OS version (server)       | Ubuntu
| Cortex Analyzer Name      | CuckooSandbox Analyzer
| Cortex Analyzer Version   | 1.0
| Cortex Version            |  2.1.3-1

### Description
The current Cuckoo Sandbox reports do not fully function with 2.0.6 Cuckoo Reports. Instead of using just hosts for the report data, we need to use domains to retrieve ip/domain combination.

### Possible Solutions
The solution is to update the analyzer and report templates to properly show the report data.

### Complementary information
I have a fix for the solution however it would also then be nice to extract the observables from the newly added ip/domain combinations.

Here is what a report currently looks with Cuckoo 2.0.6:
image

And this is after the analyzer and report fix:
image

### DIFF
--- a/analyzers/CuckooSandbox/cuckoosandbox_analyzer.py
+++ b/analyzers/CuckooSandbox/cuckoosandbox_analyzer.py
@@ -105,17 +105,17 @@ class CuckooSandboxAnalyzer(Analyzer):
             else:
                 snort_alerts = []
             try:
-                hosts = [(x['ip'], x['hostname'], x['country_name']) for x in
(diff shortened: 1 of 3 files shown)


nicpenning:
This still might not cover everything. I am re-evaluating this report/analyzer. Instead of using Domains and having a single IP it will be better to get the DNS responses since one hostname could have many IP Addresses.

image

PR #421, opened 2019-02-10
A responder for the Crowdstrike Falcon custom IOC api

ag-michael:
I've made a responder that will submit observables to the Crowdstrike Falcon custom IOC api. 
Blog post about this feature: https://www.crowdstrike.com/blog/tech-center/import-iocs-crowdstrike-falcon-host-platform-via-api/

Crowdstrike Falcon customers can use this responders to submit case observables to their IOC list maintianed by Crowdstrike. When the observables are seen by Falcon in their environment, a detection will be generated.


xg5-simon:
Consider using the operations function to add a tag to the observable to provide a clear indicator to analysts that the observable has been uploaded to CrowdStrike Falcon without having to click into the observable details view.

For example

`def operations(self, raw):
    return [self.build_operation('AddTagToArtifact', tag='CrowdStrike:IOC Uploaded')]`

saadkadhi:
The change suggested by @xg5-simon would be helpful @ag-michael. May you consider it? Thanks.

ag-michael:
@xg5-simon I've added the change you requested. Thanks for the suggestion, I read that part in the responder guide but forgot to implement it.

PR #400, opened 2019-01-03
Added AbuseIPDB analyzer

mlodic:
Related to issue #353

PR #425, opened 2019-02-14
added templates for AbuseIPDB

mlodic:
Related to #400.

Forgive me but I don't use The Hive, I'm just a Cortex user and, sincerely, I didn't get the chance to test them. However I followed other templates scheme. If you can check if it works before merging it would help. thanks


3c7:
Thank you @mlodic :)

nadouani:
Hello, I think these reports needs to be refined to display a table with a row for each finding.

jeromeleonard:
made this template if you are ok
Screen Shot 2019-03-23 at 7 08 55 PM

PR #428, opened 2019-02-18
Updated crt.sh analyzer to display SHA-1 

kx499:
hash of cert. Needed to look over results and call the id to get it. Updated the report html as well to show in a dl instead of a table. SHA1 in report links out too crt.sh detail page for the cert


kx499:
closing this pr and submitting under a different branch so i can continue working on additional updates

PR #434, opened 2019-03-01
DNSSinkhole analyzer

garanews:
Analyzer for query DNSSinkhole installed locally (ex. using BIND)

image


jeromeleonard:
Thank you @garanews for this analyzer. I see that you consider a domain is sinkholed only if the dns query returns `127.0.0.2`. If your sinkhole has another IP address, it will return `False`. Maybe an idea would be to also configure the sinkhole IP address,  and check if the domain resolves with this address instead

garanews:
I added it!

immagine

jeromeleonard:
thx you, reviewing it quickly

PR #457, opened 2019-04-10
remove builtin modules from requirements.txt

ag-michael:
Remove builtin python modules from `requirements.txt`


3c7:
Thanks @ag-michael!

PR #478, opened 2019-04-26
yeti api key

siisar:
Optional yeti api key

### DIFF
--- a/analyzers/Yeti/Yeti.json
+++ b/analyzers/Yeti/Yeti.json
@@ -15,6 +15,13 @@
       "type": "string",
       "multi": false,
       "required": true
+    },
+    {
+      "name": "api_key",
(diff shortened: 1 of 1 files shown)


jeromeleonard:
correct me if i'm wrong but the `api_key` parameter is not used in the code.

siisar:
Hi Jerome, it is true, because of that I have another pull request where I change the code to use it
https://github.com/TheHive-Project/Cortex-Analyzers/pull/477

I know it should have been both changes in the same pull request, I did it in two by mistakeRegards


    En miércoles, 8 de mayo de 2019 19:25:09 CEST, Jerome Leonard <notifications@github.com> escribió:  
 
 
correct me if i'm wrong but the api_key parameter is not used in the code.

PR #483, opened 2019-05-22
Updating to use new API

obikao:
This now uses the new HTTPS API for SinkDB, documentation is available here:
https://sinkdb.abuse.ch/api

I wasn't sure how to credit the original author, so I put the original URL in the description. Feel free to change that however you want. This also doesn't have an associated HTML template with it yet. I will be working on that next.


axpatito:
Why is this PR using 8 space tabs? Also, if you are working on top of someone else work, you cannot delimit yourself as the author and erase the original author.

obikao:
Well, on my end they are 4 space tabs. Not sure why that's showing up for you. And second, this is my first OSS submission where I ended up rewriting the entire thing. However, not knowing what to do, this is why I added the note about crediting the original author and asked the owner of the repository to change it as they see fit.

PR #486, opened 2019-05-29
EmlParser - Fixed headers and displayTo 

### LINKED ISSUE #484: [Bug] EmlParser has incomplete header
**Describe the bug**
The header show by EmlParser is incomplete.
There's only the last 'Received' information

**Expected behavior**
The header should have multiple 'Received' atributes for all the points it has passed.

**Possible solutions**
Problably the 'Received's are being lost because of the use of json (or dict), as it can have only one attribute of the same name.
Extracting the entire header as a string should solve the problem (if possible).

mgabriel-silva:
Fixed headers not showing all 'Received' information. Now it's a string
Fixed displayTo attribute name not compatible with the thehive template
Fix #484


mgabriel-silva:
Fixed the template as well

Issue #487, opened 2019-05-30
[Bug] IBM X-Force Analyzer adds an extra slash which prevents it from running correctly


saadkadhi:
**Describe the bug**
As [originally reported](TheHive-Project/Cortex#84) by @csoc49 in the Cortex repository:

```
Hy,

we have a problem by connecting to the IBM X-Force API.
We always get a 401 API error.

We use these URL https://api.xforce.ibmcloud.com/ and take our API-Key and pwd in the Cortex configuration. Is there anybody with an idea??

:-/
```

**Complementary information**
@ssstonebraker reported the following:
> Hey i figured out how to fix this. I had the same problem and spent a couple hours figuring it out

> I kept receiving an "API error" when attempting to run TheHive Project Cortex Analyzer for IBM Xforce.

> There is currently a bug that is adding an extra / with every request.

> To fix this issue you need to modify this file:

`/opt/Cortex-Analyzers/analyzers/IBMXForce/ibmxforce_lookup.py`

> Command to fix the problem:

`perl -pi -e 's|%s/|%s|g' /opt/Cortex-Analyzers/analyzers/IBMXForce/ibmxforce_lookup.py`

> source:

https://brakertech.com/thehive-project-cortex-ibm-xforce-analyzer-is-not-working/

ayounes2024:
hi team, i run the command but still issue exist

PR #488, opened 2019-05-30
remove extra slash

garanews:
it fixes https://github.com/TheHive-Project/Cortex-Analyzers/issues/487
Thanks @csoc49 for reporting
Thanks @ssstonebraker for found fix

At the time analyzer was written that slash was needed :)

### DIFF
--- a/analyzers/IBMXForce/README
+++ b/analyzers/IBMXForce/README
@@ -10,9 +10,6 @@ IBMXForce {
        }
 ```
 
-To test the analyzer from cmdline
-
-python IBMXForce_lookup.py < input
 
 Testing
 --------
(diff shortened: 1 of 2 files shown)


jeromeleonard:
Currently reviewing. I don't have access to this service. I assume that this is due to a change of the URL to access to the service ? (It has now a trailing '/' that it hasn't before) ?
 
Also in the fix: 

line 122 `_session.get('%/sresolve/%s'` 

Could you confirm that it is `_session.get('%sresolve/%s'` instead ? 

If yes I change it before merging.

garanews:
I just tested again and now every request need the slash... so the original code works!
I presume someone added url with slash at the end, going to strip it.

jeromeleonard:
thx @garanews, that is what I supposed while reviewing the code. It is better like this.

PR #498, opened 2019-06-16
[WIP] SinkDB: fix multiple issues

ilyaglow:
* Rename SinkDB 3rd level domain that IP is resolved against.
* Fix resulting IP that SinkDB returns for a sinkholed IP.
* Handle additional categories: scanner, phishing awareness.

### DIFF
--- a/analyzers/SinkDB/sinkdb.py
+++ b/analyzers/SinkDB/sinkdb.py
@@ -17,30 +17,39 @@ class SinkDBAnalyzer(Analyzer):
         self.data = '.'.join(self.data)
 
     def dig(self, ip):
-        proc = subprocess.Popen(['dig', '+short', '{}.{}.sinkdb-api.abuse.ch'.format(ip, self.apikey)],
+        proc = subprocess.Popen(['dig', '+short', '{}.{}.sinkdb-dnsapi.abuse.ch'.format(ip, self.apikey)],
                                 stdout=subprocess.PIPE,
                                 stderr=subprocess.PIPE)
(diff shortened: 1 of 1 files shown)


jeromeleonard:
Hello @ilyaglow, 

This PR is tagged [WIP]. Do you consider it ready to be merged ?

Thx.

dadokkio:
Hi @jeromeleonard, this pull and #483 are both supporting v2 sinkdb api.
In my opinion #483 is done in a better way, so we can close this and keep the other one.
I've also updated #483 to support fqdn and url in pull #756.

ilyaglow:
Hello, this PR doesn’t have a template changes for TheHive, but it works fine for me. At the time of submitting, I probably didn’t notice #483. However, this analyzer can run without API-key, compared to #483 and #756. 

@dadokkio maybe we can incorporate in #483 DNS lookups as a fallback in case an API key is not specified?

dadokkio:
Yes, we could but I've some doubts (because I know dig just a little :P )
The API version works also for fqdn, domain and mail; this way will work just for ip?
In the code I see that also in your case the api key is attached to the query, in case of no api key it will lookup for ***.None.sinkdb-dnsapi.abuse.ch ?

ilyaglow:
Whoops! You're right. It still needs the API key for DNS lookups. Sorry for being sloppy :)

So yeah, this PR can be closed.

dadokkio:
No problem, thank you for the pull in any case :smile:

PR #499, opened 2019-06-16
FileInfo: fix ioc_parser package in requirements

ilyaglow:
Package `iocp` in requirements.txt [refers to a wrong package](https://pypi.org/project/iocp/).

We should use `ioc_parser` instead. The official one [looks unmaintained](https://github.com/armbues/ioc_parser) anymore, so I added a @SteveClement's [fork](https://github.com/SteveClement/ioc_parser).


jeromeleonard:
fixed

Issue #508, opened 2019-07-03
[Bug]Missing baseConfig in two Analyzsers


github-pba:
**Describe the bug**
The two analyzers
* URLhaus
* Urlscan.io

lack of the baseConfig parameter in their service interaction files (*.json) and cant be activated for that reason.

Original finding by [b00lpy](https://github.com/b00lpy) in [cortex issue 200](https://github.com/TheHive-Project/Cortex/issues/200)

**To Reproduce**
1. cd /path/to/analyzers
1. grep -c baseConfig */*.json 

Those lines with a zero count lack of the parameter

**Work environment**
Analyzers cloned on July 3rd, 2019

**Possible solutions**
Add the baseConfig parameter to JSON files.

PR #511, opened 2019-07-09
#508 fix missing baseConfig param

github-pba:
Fix for issue 508: missing baseConfig param in *.json.

catalog rebuilt.


jeromeleonard:
This has already been fixed.

PR #517, opened 2019-07-26
removed python builtins from requirements.txt

github-pba:
Removed python builtins from `responders/FalconCustomIOC/requirements.txt`. These throw error messages during install with `pip install -U -r requirements.txt`.

Fix for issue 509


3c7:
Thanks @github-pba.

PR #525, opened 2019-08-20
[BugFix] HIBP Analyser no longer works

### LINKED ISSUE #524: [Bug] HIBP Analyser no longer works
**Describe the bug**
On July 2019 they updated the API to use [authentication](https://haveibeenpwned.com/API/v3#Authorisation). All queries will now fail.

**To Reproduce**
1. Run the HIBP analyser 
1. ???
1. Profit
1. Nope.

**Expected behavior**
A valid response from the API. 

**Possible solutions**
We need to update the analyser to take an API key in the configuration. 

There will be changes to the analyser code:
1. Change to the v3 API
1. Adding in authentication header

jonashergenhahn:
Changed the analyser to work with new api version 3. 

New Configuration options added:
- api_key: API key needed for queries (required)
- truncate: choose if you want truncate responses (required, default false)
- retries: max tries while getting 429 status codes (too many requests, see also [RateLimiting](https://haveibeenpwned.com/API/v3#RateLimiting)) (optional)

Changed thehive template:
- modified for truncate responses

Closes #524


jeromeleonard:
@jonashergenhahn thank you for the PR.

PR #530, opened 2019-09-05
Csoauth2

ag-michael:
Crowdstrike has changed how their API authentication works, they are now using OAUTH2 and have deprecated the legacy authentication API. 

I have added a new responder `Crowdstrike Falcon Custom IOC API OAUTH2_2_0`  which runs on python3 (old responder was python2). 

While creating this, with the latest cortexutils I ran into a bug where the responder (via cortexutils) was expecting input from stdin but cortex was saving it to /tmp/<job folder>/input/input/json: 
https://github.com/TheHive-Project/Cortex/blob/13ecd332cd63becdc98b470ae989425b3feb7a43/app/org/thp/cortex/services/JobRunnerSrv.scala#L93

The new responder 'remaps' stdin to this file:

```
def cortexinputbug():
    import sys
    if len(sys.argv) > 1:
        try:
            sys.stdin = open(sys.argv[1]+"/input/input.json")
        except:
            with open("/tmp/responder_error", "w+") as e:
                e.write(traceback.format_exc())
```
please feel free to remove this work around if the bug in cortexutils has been fixed.

### DIFF
--- a/responders/FalconCustomIOC/FalconAuth.py
+++ b/responders/FalconCustomIOC/FalconAuth.py
@@ -1,0 +1,33 @@
+import time
+import json
+import requests
+
+
+class FalconAuth:
+    def __init__(self, client_id, client_secret):
+        self.client_id = client_id
+        self.client_secret = client_secret
+        with open("/tmp/OAuth2.json", "w+") as f:
+            f.write("")
+
+    def newtoken(self):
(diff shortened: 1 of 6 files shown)


nusantara-self:
Hello, closing the PR as it is now outdated & some set of CSFalcon analyzers & responders are now supported (see [3.4.0](https://strangebee.com/blog/cortex-analyzers-3-4-0-unleashing-the-falcon/))

Cheers,

PR #538, opened 2019-10-03
Fix - updated cortexutil Extractor return keys

dadokkio:
CortexUtil changed Extractor return keys on 4th April

https://github.com/TheHive-Project/cortexutils/commit/f63222df1744c40e17becc2719f1337c5251e6d5#diff-8bfea30ee07c21ace8c48037f4b8457e


garanews:
Will add this in milestone 2.4?

nadouani:
Sounds like necessary yes :)

Issue #521, opened 2019-08-09
Talos Analyzer No Longer Works


colin-stubbs:
**Describe the bug**
Talos now uses rolling cookies on their API, requests to API without a valid cookie do not work.

**To Reproduce**
Use existing analyzer, it is no workies.

**Expected behavior**
Works.

**Complementary information**
n/a

**Work environment**
 Latest

**Possible solutions**
Need to use requests.Session() against https://talosintelligence.com/reputation_center/lookup?search=IPADDRESS
This obtains cookies, then subsequent requests to API work

**Additional context**
nOpe.

robertnixon2003:
Lets get this merged. I use this one alot.

PR #546, opened 2019-10-18
fixed Talos analyzer

0xmilkmix:
fixed issue #521  by changing the user-agent, accepted format and encoding


robertnixon2003:
Can we get this merged?

0xmilkmix:
> Can we get this merged?

checks are failing but I don't understand why. Looks like the error might be outside of the code

3c7:
> > Can we get this merged?
>
> checks are failing but I don't understand why. Looks like the error might be outside of the code

Indeed, the failing check has nothing to do with your code.

PR #559, opened 2019-11-01
update of long.html to add site screenshot

anscz:
Change adding screenshot thumb of site directly to the report, you can click on it for full size screenshot. If there are more results screenshot is added under each occurrence.

### DIFF
--- a/thehive-templates/Urlscan_io_Search_0_1_0/long.html
+++ b/thehive-templates/Urlscan_io_Search_0_1_0/long.html
@@ -8,18 +8,28 @@
     </div>
     <div class="panel-body">
         <table class="table table-hover">
-            <tr>
-                <th>URL</th>
-                <th>Result</th>
-                <th>Time</th>
-            </tr>
             <tr ng-repeat="r in content.indicator.results | limitTo:results_limit | orderBy:'-task.time'">
-                <td>{{r.page.url | ellipsis:40}}</td>
-                <td>
-                    <a href="https://urlscan.io/result/{{r._id}}" target="_blank">https://urlscan.io/result/{{r._id}}</a>
-                </td>
-                <td>{{r.task.time}}</td>
-            </tr>
+				<td>
+					<table class="table table-hover">
+                        <tr>
+                            <th>URL</th>
+                            <th>Result</th>
+                            <th>Time</th>
+                        </tr>
+						<tr>
+							<td>{{r.page.url | ellipsis:40}}</td>
+							<td>
+								<a href="https://urlscan.io/result/{{r._id}}" target="_blank">https://urlscan.io/result/{{r._id}}</a>
+							</td>
+							<td>{{r.task.time}}</td>
+						</tr>
+						<tr>
+							<td colspan="3" style=text-align:center>
+								<a href="https://urlscan.io/screenshots/{{r._id}}.png"><img src="https://urlscan.io/thumbs/{{r._id}}.png"></a>
+							</td>
+						</tr>
+					</table>
+				</td>
         </table>
     </div>
 </div>


nadouani:
Hello, thanks for the PR. IMHO, I don't think it's a good idea the have TheHive refer to external links like at `urlscan.io` this might disclose some internal info. 
Any thoughts

3c7:
Yep, also thought about that.

anscz:
Hi, in which way? As you get that only when you click on generate report (pass all TLP / PAP checks) and at that time those data are already available at urlscan.io there, on top of it it's screenshot publicly available webpage. Can you please elaborate more on your concern? Btw many reports refer to external sites anyway.

nadouani:
@anscz the `` will automatically call urlscan.io with your internal TheHive domain as referrer. I don't know if this is a good idea.

@3c7 are you thinking about something else?

anscz:
@nadouani  ach ok, I thought disclosure in other meaning, this definitely make sense. I'm trying to use:
`<meta name="referrer" content="never">
<meta name="referrer" content="no-referrer">`

but with unreliable results where EDGE ignores it completely, while FF partially and only in chrome works well :/ .

nadouani:
It sounds like the `img` tag has a `referrerPolicy` attribute:

https://developer.mozilla.org/en-US/docs/Web/API/HTMLImageElement/referrerPolicy

anscz:
True I try that as well, right now I'm little bit lost because browser tries to load two png instead of one. can you help me to understand how that iteration in report template works.
important part of template is:
`    <div class="panel-body">
        <table class="table table-hover">
            <tr ng-repeat="r in content.indicator.results | limitTo:results_limit | orderBy:'-task.time'">
				<td>
					<table class="table table-hover">
                        <tr>
                            <th>URL</th>
                            <th>Result</th>
                            <th>Time</th>
                        </tr>
						<tr>
							<td>{{r.page.url | ellipsis:40}}</td>
							<td>
								<a href="https://urlscan.io/result/{{r._id}}" referrerpolicy="no-referrer" target="_blank">https://urlscan.io/result/{{r._id}}</a>
							</td>
							<td>{{r.task.time}}</td>
						</tr>
						<tr>
							<td colspan="3" style=text-align:center>
								<a href="https://urlscan.io/screenshots/{{r._id}}.png" referrerpolicy="no-referrer"></a>
							</td>
						</tr>
					</table>
				</td>
        </table>`
there are two get requests
**one for "untranslated" {{r._id}} which contains as well referrer**
`Request URL: https://urlscan.io/thumbs/%7B%7Br._id%7D%7D.png
Request Method: GET
Status Code: 200  (from disk cache)
Remote Address: 148.251.45.170:443
Referrer Policy: no-referrer-when-downgrade
Provisional headers are shown
Referer: https://10.193.65.100:9443/index.html
User-Agent: Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/78.0.3904.108 Safari/537.36`

**and another which is as should be**
`Request URL: https://urlscan.io/thumbs/146bcd95-0fb9-4687-b486-5a417bc3d3bb.png
Request Method: GET
Status Code: 200  (from disk cache)
Remote Address: 148.251.45.170:443
Referrer Policy: no-referrer
Provisional headers are shown
User-Agent: Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/78.0.3904.108 Safari/537.36`

anscz:
I did some tests and looks like that order of atribute matter. Now, it's tested with FF and Chrome as well as latest EDGE fully supports referrerpolicy="no-referrer" atribute. I edited original PR file.
S

PR #568, opened 2019-12-09
Simple analyzer for GRR

pettai:
Simple analyzer for GRR that checks and map the GRR client_id of an IP / FQDN

### DIFF
--- a/analyzers/GRR/GRR.json
+++ b/analyzers/GRR/GRR.json
@@ -1,0 +1,37 @@
+{
+    "name": "GRR",
+    "version": "0.1",
+    "author": "pettai@sunet.se, SUNET",
+    "url": "https://github.com/TheHive-Project/Cortex-Analyzers",
+    "license": "AGPL-V3",
+    "description": "Search GRR for the host agent.",
+    "dataTypeList": ["ip", "fqdn"],
+    "baseConfig": "GRR",
+    "config": {
+       "service": "query"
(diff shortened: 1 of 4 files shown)


dadokkio:
I tried the analyzer and even if it's working it's not clear for me how it can be used.
I've installed grr and I've added some clients running random task on them.
Looking for a client fqdn it returns grr uuid, and just that. 
It's the only information you need? Because probably I'm missing something :smiley:

pettai:
hi @dadokkio 
Yes, it's that simple, it return the UUID if the client is under GRR control. (That's typically what Analysers do, "do I know anything about this observable (IP or FQDN in this case)).

I'm working on a Responder for GRR, here is where the more useful things happens, like hunting for artifacts. But In order to run the Responder, you need to know if this client is under GRR-control and which UUID it has...

dadokkio:
Ok, my idea was to add missing templates and I wanted to be sure I wasn't missing something.
The tool has a lot of functionality but you're right responder is a better place them.
So, I'm going to add templates to your pull so we'll be able to merge it.

dadokkio:
I updated your pull with thehive templates and a notfound taxonomy https://github.com/TheHive-Project/Cortex-Analyzers/commit/f0737d0e6dd92c8a16203cf803c26fa5afe6012e

image

image

pettai:
ok thx

PR #467, opened 2019-04-19
add Emailrep analyzer

ninoseki:
Implement #466

PR #575, opened 2019-12-19
Update EmailRep analyzer

ninoseki:
Catch up the latest API response schema (Ref. #467)


3c7:
That was fast! Thank you @ninoseki!

PR #589, opened 2020-01-07
LdapQuery Analyzer

cyberpescadito:
Hello,

Today I wanna share a simple LdapQuery analyzer. It's useful to get the context of an user from your organization for investigations purposes.

Configure it:
- Configure the analyzer on Cortex (ldap server address, account to bind, attributes to fetch)
- Add the report templates to your Hive

Run in TheHive: 
-Declare an observable (accepted types: username & email address)
-Run the Analyzer


nadouani:
Hello @cyberpescadito  Thanks for the PR, is it possible to add a PR for https://github.com/TheHive-Project/CortexDocs/blob/master/analyzer_requirements.md

analyzers/LdapQuery/LdapQuery.json
```diff
@@ -0,0 +1,48 @@
...
+  "author": "Florian Perret @cyber_pescadito",
+  "url": "https://github.com/cyberpescadito/Cortex-Analyzers/tree/master/analyzers/LdapQuery",
+  "license": "AGPL-V3",
+  "description": "Query your LDAP server to harvest informations about an user of your organization",
+  "dataTypeList": ["username", "mail"],
+  "command": "LdapQuery/LdapQuery.py",
+  "baseConfig": "LdapQuery",
+  "configurationItems": [
+    {
+        "name": "LDAP Server Address",
```
nadouani: I would recommend using the "Snake Case" notation: `xxx_yyyy_zzzz` for all the configuration items' name

analyzers/LdapQuery/LdapQuery.py
```diff
@@ -0,0 +1,74 @@
...
+        self.attributes = self.get_param('config.Attributes', None, 'Missing attributes list to report')
+        self.payload = self.get_param('data', None, 'username to search in LDAP is missing')
+
+    def summary(self,raw):
+        taxonomies = []
+        level = "info"
+        namespace = "LDAP"
+        predicate = "Query"
+        value = "Success"
+        taxonomies.append(self.build_taxonomy(level, namespace, predicate, value))
```
nadouani: Is there any useful information you can put as the value of that taxonomie? probably the `cn` or the `uid`?

analyzers/LdapQuery/LdapQuery.py
```diff
@@ -0,0 +1,74 @@
+#!/usr/bin/env python
+#Author: @cyber_pescadito
+import ldap
+import json
+from cortexutils.analyzer import Analyzer
+
+class LdapQuery(Analyzer):
+    def __init__(self):
+        Analyzer.__init__(self)
+        self.ldap_address = self.get_param('config.LDAP Server_Address', None, 'ldap_address is missing')
```
nadouani: missing `_` in `LDAP Server_Address`

analyzers/LdapQuery/LdapQuery.py
cyberpescadito: I agree it would be strange to not include "cn" value in attributes list to fetch from the LDAP, but if "cn" isn't declared in "attributes" list by the admin in cortex analyzer configuration it could break the summary. So i chosen to not include any value of LDAP there.
What is your point of view? should I change de program design to enforce "cn" attribute fetching ?

nadouani: No, the idea was to make that taxonomie meaningful/useful. 

If it will be always: `Ldap:Query=Success` then it's not worth it.

I would make a taxonomy like: `LdapQuery:cn=XXX` or `LdapQuery:ATTRIBUTE_NAME=ARRTIBUTE_VALUE` or nothing

analyzers/LdapQuery/long.html
```diff
@@ -0,0 +1,38 @@
+<div class="report-FILEInfo" ng-if="success">
+    <style>
+        .report-FILEInfo dl {
```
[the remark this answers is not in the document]
nadouani: Juste replace `.report-FILEInfo` by `.report-LdapQuery`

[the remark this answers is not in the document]
cyberpescadito: Too much inspired from my model there :D thanks

analyzers/LdapQuery/LdapQuery.py
cyberpescadito: is value a mandatory attribute of taxonomies?

nadouani: Note sure, but the taxonomy aims to provide important info, without having to read the full report. You can probably show a taxonomy for important attributes if they exist?

dadokkio:
I updated your pull following our discussion.
Changes:
- use ldap3 instead of python-ldap
- add search field to support Active Directory
- add port settings to support ldaps
- template now show all fields that you decide to return

PR #603, opened 2020-01-30
censys search

CTIMarc:
Adds censys search on ipv4 dataset

### DIFF
--- a/analyzers/Censys/Censys_search_ipv4.json
+++ b/analyzers/Censys/Censys_search_ipv4.json
@@ -1,0 +1,27 @@
+{
+  "name": "Censys_search",
+  "author": "Marc Nebout, SEKOIA",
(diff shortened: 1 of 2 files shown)


jeromeleonard:
Hello @CTIMarc, is there a difference with existing Censys analyzer ?

CTIMarc:
Yes, it adds the possibilty to search anything on IPV4 dataset. The existing one doesn't permit to search multiple criterias.

dadokkio:
Some thoughts:
- there are parameters in __init__ that are not available in json, so it'll take default everytime.
- the code is very similar to the existing one, probably is better to have just one python file for both and integrate your changes there

dadokkio:
Closing since new pull with code merged instead of new analyzer.

Issue #361, opened 2018-10-22
Force python3 in all analyzers


jeromeleonard:
Review all analyzer and make them working with python3 (at leat, try to :)  )

List of analyzers that might be reviewed : 

- [x] Abuse_Finder
- [x] StopForumSpam
- [x] MaxMind
- [x] Virustotal

### Doc

- [ ] Update "How to create an analyzer" 

**To be completed**

LetMeR00t:
Hi guys,
I proposed this PR for Abuse_Finder but it will need another PR (in the repository abuse_finder) to be used :
https://github.com/TheHive-Project/Cortex-Analyzers/pull/469

nadouani:
Same thing for VirusTotal, the PR #446 solves the issue. We will merge these ASAP

nadouani:
We need to UP this one

nadouani:
@3c7 @jeromeleonard Do we need to update the the shebang statement of all the analyzers to replace `python` with `python3`?

PR #604, opened 2020-02-05
Force Python 3 usage for MsgParser analyzer

iwitz:
MsgParser works with Python 3 and doesn't work with Python2 in our environment ; this PR forces the use of Python 3. Python 2 isn't maintained since January 1st.


dadokkio:
#604 https://github.com/TheHive-Project/Cortex-Analyzers/issues/361

nadouani:
Hello @dadokkio, is this PR supposed to be included in 2.4.0 release? since it's related to #361 ?

I would say yes.

PR #607, opened 2020-02-08
Dockerize Every Analyzer

milesflo:
Addresses: https://github.com/TheHive-Project/Cortex-Analyzers/issues/606

# Changes:

1. Convert analyser JSON file versions from loose versioning schema to [semver standard](https://semver.org/)
    - This will pay dividends w/ specific analyzer version management & Docker deployment

2. Creation of 2 helper scripts for Dockerfile mgmt
    1. [`utils/dockerfile_builder.py`](https://github.com/TheHive-Project/Cortex-Analyzers/compare/master...milesflo:master#diff-26db243a3742a7e627a2f7e64352058a)
        - Utilizing the relevant config file(s) in each Analyzer, create a Dockerfile w/ proper
            - Base image (either hardcoded w/ the key `baseImage` in config or derived from source code shebang)
            - Entrypoint (derived from config `command` key)
            - Dependency installation
            - Metadata (description, author, name)
    2. [`utils/build_analyzers.py`](https://github.com/TheHive-Project/Cortex-Analyzers/compare/master...milesflo:master#diff-e64233d333432669a18a9eb27db3a66c)
        - Iterates through the Analyzers directory and builds Images for later Cortex execution

3. Inclusion of built `Dockerfile`s

4. Miscellaneous house keeping
    - Some shebangs were incorrectly defined
    - `git` pip sources should be used as a last resource in favor of PyPi
    - Updated catalogue via update_catalogue.sh
    - AutoFocus was incorrectly stylized throughout the project

# Next Steps:
1. Include these scripts in the pre-commit hook to ensure that Dockerfiles are always up to date
2. Deploy these using the [`docker image push` command](https://docs.docker.com/engine/reference/commandline/image_push/) on each of these built images at every **major release**, utilizing the [catalog system](https://github.com/TheHive-Project/Cortex-Analyzers/blob/master/analyzers/catalog.json) already in place to act as a `dockerImage` name directory.

# Known Issues:

Currently, some of the Analyzers have niche dependencies that break the automated Dockerfile generation logic... We can either change the builder script, or redesign the analyzers to use more native APIs.

## ~~Broken Analyzers:~~

- [x] Yara
    - ~~gcc does not ship with `python:2-alpine`~~
    - **Resolved** w/ dependency analysis
- [x] Malpedia
    - ~~gcc does not ship with `python:3-alpine`~~
    - **Resolved** w/ dependency analysis
- [x] EmlParser
    - ~~gcc does not ship with `python:3-alpine`~~
[...]

PR #696, opened 2020-02-12
DNSDB Python3 Support

milesflo:
These are the changes ported over from #607 that allow DNSDB to work on Python3. 

Works on my machine but please test extensively as the vendor lib was upgraded by https://github.com/guyddr/dnsdb-query (a 3rd party).

### DIFF
--- a/analyzers/DNSDB/Dockerfile
+++ b/analyzers/DNSDB/Dockerfile
@@ -1,4 +1,4 @@
-FROM python:2
+FROM python:3
 
 WORKDIR /worker
 COPY . DNSDB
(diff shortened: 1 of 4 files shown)


dadokkio:
Hi, I'm trying your analyzer but I don't know why my apy key [community edition trial] is working only few times and then returns error.

In any case when api fails it returns "Error: Bad API key" and then the analyzer fails with the following error [at line 119 when it tries to parse the error as json]:
image

Can you please check if the response is 200 before trying to parse it?

dadokkio:
Ok, it wasn't the not 200 response but all the responses returns error.

I think the problem is:
```

            self.report({
                "records": map(lambda r: self.update_date('time_first', self.update_date('time_last', r)),
                               self.execute_dnsdb_service(client))
            })
```
because map is an iterable and cannot be serialized.
If I change add list to the code in this way:
```

            self.report({
                "records": list(map(lambda r: self.update_date('time_first', self.update_date('time_last', r)),
                               self.execute_dnsdb_service(client)))
            })
```
it seems to be working.

garanews:
merged the #745

PR #697, opened 2020-02-12
Add CyberChef analyzer

weslambert:


analyzers/CyberChef/requirements.txt
```diff
@@ -0,0 +1 @@
+dnyspython
```
milesflo: Can you add cortexutils to these requirements?

weslambert: Sure thing -- it's been added.

jeromeleonard:
Hello @weslambert, 

Thank you for your PR, this is really great to be able to request CyberChef from TheHive and Cortex. Just to let you know that we updated report templates to ensure short reports won't contain large input and output that could not be displayed well in the applications. I also updated the long report so that it anyone quickly access both complete input and outputs. 

image

image

We will release it soon. Thank you again.

PR #702, opened 2020-02-21
moving to AbuseIPDB API v2

shortstack:
API v1 has been sunset, according to the responses from the API, updated to use v2

### DIFF
--- a/analyzers/AbuseIPDB/abuseipdb.py
+++ b/analyzers/AbuseIPDB/abuseipdb.py
@@ -64,7 +64,10 @@ class AbuseIPDBAnalyzer(Analyzer):
                             categories_strings = []
                             for category in item["categories"]:
                                 categories_strings.append(self.extract_abuse_ipdb_category(category))
-                        response['categories_strings'] = categories_strings
+                        try:
+                            response['categories_strings'] = categories_strings
+                        except:
+                            pass
 
                 self.report({'values': response_list})
             else:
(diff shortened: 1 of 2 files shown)


shortstack:
question--do the docker images for docker-enabled cortex pull from this repo when they're deployed? in this case, we can't test analyzer changes locally if docker analyzers are enabled :( unless i'm missing something

going back to a `process` analyzer (vs `docker`) in application.conf requires refreshing/disabling/reenabling analyzers and still seems to throw json errors when i try to go back to non-docker and run analyzers

dadokkio:
Hi, I wasn't aware of your pull request so I made a similar one with a different approach.
In my case I removed the reports information and I've just counted categories and submitter nationalities. 
This is an example view of my approach:
image

Not sure if my approach is ok, we didn't use abuseipdb so I tried to collect from hundreds of report  information I found useful.

Meanwhile, I've also tested your code and it's working properly. 
The only missing piece are templates for thehive. Are you able to add those to your pull?

shortstack:
hi @dadokkio,

thanks for responding. i updated it only to do what it was already doing, but with the new API configuration. regarding the report template, i will add. thank you! :)

dadokkio:
Perfect, I just made one small fix https://github.com/TheHive-Project/Cortex-Analyzers/commit/944afb608d36dca91947861d60a0fd1b5ac594f8
to render categories name in template.
image
If that is ok I'm going to merge your pull.

shortstack:
awesome! thank you!! :)

dadokkio:
merged with #719

PR #721, opened 2020-03-19
Revert "Bumped Investigate version"

jeromeleonard:
Reverts TheHive-Project/Cortex-Analyzers#718


jeromeleonard:
Created a trial account on Cisco Investigate, tested with version fixed  earlier and docker image and it works. no need to bump the version.

Issue #723, opened 2020-03-20
New Analyzer: OpenCTI


amr-cossi:
### Request Type
Analyzer

### Description
This Analyzer allows to search for an observable in several instances of OpenCTI and returns details about this observable and all reports linked to it.
For the first version, it comes only in one flavor to look for any type of observable.

garanews:
available in 2.7.0

PR #725, opened 2020-03-20
Add OpenCTI Analyzer v1

amr-cossi:
https://github.com/TheHive-Project/Cortex-Analyzers/issues/723

### DIFF
--- a/analyzers/OpenCTI/OpenCTI.json
+++ b/analyzers/OpenCTI/OpenCTI.json
@@ -1,0 +1,57 @@
+{
+  "name": "OpenCTI_SearchObservable",
+  "author": "ANSSI",
+  "license": "AGPL-V3",
+  "url": "https://github.com/TheHive-Project/Cortex-Analyzers/",
+  "version": "1.0",
+  "description": "Query multiple OpenCTI instances for an observable.",
+  "dataTypeList": [
+    "domain",
+    "ip",
+    "url",
+    "fqdn",
+    "uri_path",
+    "user-agent",
+    "hash",
+    "email",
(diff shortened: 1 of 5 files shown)


dadokkio:
Just little issues in template. 
All the if are based on .colour prop instead of .color so all background changes are failing:
image

Also externak refs has label class so it's written white on white.

amr-cossi:
Thank you @dadokkio for the test and the feedback. I admit that I don't use TheHive templates in my main use cases so I tried to provide this one based on existing ones but without proper testing.
Please let me know if you see any more bug or any improvement that can be done for TheHive users.

dadokkio:
@amr-cossi we were thinking about writing a responder in order to push infomation back from thehive to openCTI [maybe related to https://github.com/OpenCTI-Platform/connectors/issues/3 ] 
you have some suggestion? your client-python library can be used for this?

amr-cossi:
I would think of 2 possible ways to achieve this :
 - a Cortex Responder using the same official Pycti library than this Analyzer. This API wrapper lets you both query data and add new data (see https://github.com/OpenCTI-Platform/client-python/tree/master/examples)
 - an OpenCTI connector periodically querying TheHive API or database to add relevant data (full mirror or filter on a specific field)

For the other way around, this Analyzer is for IOC searching only. A use case where a TheHive case should be opened from OpenCTI could also exist. In https://github.com/OpenCTI-Platform/connectors/issues/3, Samuel suggested that the OpenCTI connector could be bi-directionnal. It could be able call the Cortex responder to create a TheHive case based on some filtering.

PR #730, opened 2020-03-26
Fix bug emlparser when 'content-type' string in mail is in lower case

TofBaasken:
Emlparser crashes when the string "Content-Type" is not found. In some cases it's available as "Content-type" or "content-type" , this fixes those cases


garanews:
merged in 2.7.0

PR #397, opened 2018-12-21
New analyzer: Tenable SecurityCenter

InsanesTheName:
This analyzer launches a scan via SecurityCenter on an IP/FQDN using the new SecurityCenter API.

PR #758, opened 2020-05-07
New analyzer: Tenable SecurityCenter

InsanesTheName:
This analyzer launches a scan via SecurityCenter on an IP/FQDN using the new SecurityCenter API.

Resubmission of #397 per request.


InsanesTheName:
Hi @jeromeleonard - resubmitted for you.

Unfortunately I will need help creating templates. I've left the role I was in since I developed this, and I no longer have an environment to use/test with.

Issue #765, opened 2020-05-13
New Responder: Virustotal Downloader


hariomenkel:
I created a responder which takes a hash as input, downloads it from Virustotal (If it exists) and adds the downloaded file back to the parent case. If a file extension could be guessed from file magic bytes the extension is added to the new observable.

https://github.com/hariomenkel/VirustotalDownloader

PR #766, opened 2020-05-13
Add new responder VirustotalDownloader #765

hariomenkel:


jeromeleonard:
This responder is using thehive4py lib to connect to the API of TheHive. Having responders connecting to TheHive to run operations in it is not the right way do do. Cortexutils has a "operation" method that allows the responder to do some actions inside TheHive when receiving the report, like _AddTagToCase_, _AddTagToArtifact,_ and **_AddArtifactToCase_**. 
Unfortunately, This is not handling file dataType yet. ( https://github.com/TheHive-Project/TheHive/issues/1413 ).

A future version of the program could be without using TheHive API, and even be included inside in existing VirusTotal Analyzer as soon as Cortex and TheHive support operations for Analyzers ( https://github.com/TheHive-Project/TheHive/issues/1406 )

PR #767, opened 2020-05-18
Fix bug in PhisTank analyzer. JSON was loading bytes instead of string.

mamoedo:
A TypeError exception is raised when the analyzer is run.

`TypeError: the JSON object must be str, not 'bytes'`

**json.loads** function requires a string, and the content of the requests response is bytes. Changing _content_ to _text_ fixes the bug.


dadokkio:
Closed as duplicated.

PR #772, opened 2020-05-22
Added VULNS taxonomies from Shodan and flag suspicious tag

CERT-ARKEA:
Add vulns (CVE) list from host, tag shodan: CVE:xxxx whith suspicious level color

### DIFF
--- a/analyzers/Shodan/shodan_analyzer.py
+++ b/analyzers/Shodan/shodan_analyzer.py
@@ -46,6 +46,7 @@ class ShodanAnalyzer(Analyzer):
     def summary(self, raw):
         taxonomies = []
         level = "info"
+        levelorange = "suspicious"
         namespace = "Shodan"
         predicate = "Location"
         if self.service in ['host', 'host_history']:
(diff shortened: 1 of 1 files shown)


dadokkio:
Hi, tried with random samples on shodan but some of them have too much vulns in them

image

Probably is better to show them only if there are few vulns and count them in the other case.

CERT-ARKEA:
Hi,
add modification, now count CVE result and return only CVE names if result < 3, else return count number.

dadokkio:
OT. you think it'll be useful to add those info also in the long template?

CERT-ARKEA:
hum, for me, the CVE numbers that affect the observable is important. This numbers is in the shodan json return by cortex. I'm not convinced that the count of CVE is useful in this template (you speak well about de json return of Shodan host analyser ?) but it's my own opinion :) Perhaps other users need this information to determinate the criticity of observable

dadokkio:
the question was because with your fix I can see the cve info in taxonomies but when I open the observable to check the long template that info is not present. Probably we could add a cve panel with the list of cve and links to their definitions

CERT-ARKEA:
oh yes that's a great idea, the CVE panel will be extremely interessant

dadokkio:
This is a first attempt:
image

CERT-ARKEA:
Nice job ! Very useful indeed :)

dadokkio:
I tried to add to templates to your pull and I ended up with a new one #776

nadouani:
This PR has been merged manually. Thanks @CERT-ARKEA

PR #784, opened 2020-06-08
VMRay: fix comparison of list and int instead of len and int

mback2k:
Fix exception happening during taxonomy creation, unfortunately hidden by cortexutils.


dadokkio:

PR #785, opened 2020-06-08
VMRay: fix IndexError: list index out of range on resubmit

mback2k:
Calling self.report does not stop execution, return is needed.


mback2k:
The IndexError would otherwise happen in this line, because on resubmit there are no submissions:
https://github.com/TheHive-Project/Cortex-Analyzers/blob/94f22129cf9d6d1a703b2446307cf94f6d2c268f/analyzers/VMRay/vmray.py#L51

dadokkio:
I tested both pulls and everything is ok :+1:

dadokkio:

PR #786, opened 2020-06-08
PhishTank: fix TypeError: the JSON object must be str, not 'bytes'

mback2k:
Use the requests built-in json method instead of json.loads.


dadokkio:
Duplicated #767

mback2k:
> Duplicated #767

Sorry, but I think my fix is the better one in this case, because it uses the built-in method.

dadokkio:

Issue #805, opened 2020-06-26
[Bug] CIRCLPassiveSSL uncaught exception on unexpected server behavior


0xmilkmix:
**Describe the bug**

Retrieving pssl information for an IP always end-up with `AttributeError: 'NoneType' object has no attribute 'get'` in https://github.com/TheHive-Project/Cortex-Analyzers/blob/master/analyzers/CIRCLPassiveSSL/circl_passivessl.py#L42

**To Reproduce**

Either running an analyzer against `8.8.8.8` or reproducing its actions in python:

```
import pypssl
ip = '8.8.8.8'
pssl = pypssl.PyPSSL(basic_auth=(..., ...) )
result = pssl.query(ip)
certificates = list(result.get(ip).get('certificates'))
for cert in certificates:
  print({'fingerprint': cert, 'subject': result.get(ip).get('subjects').get(cert).get('values')[0]})
{'subject': 'C=US, O=Google Inc, CN=Google Internet Authority G2', 'fingerprint': 'd6ad07....87948'}
{'subject': 'C=US, ST=California, L=Mountain View, O=Google LLC, CN=dns.google', 'fingerprint': 'e7ca394...baf5'}
{'subject': 'C=US, ST=California, L=Mountain View, O=Google Inc, CN=*.c.docs.google.com', 'fingerprint': '2b1cfcc...94e'}
Traceback (most recent call last):
  File "<stdin>", line 2, in <module>
AttributeError: 'NoneType' object has no attribute 'get'
```

**Expected behavior**

From debugging session, `result.get(ip).get('certificates')` contains more hashes than entries in `result.get(ip).get('subjects')`.

For example in the example above, `dfe2070c79e....9c2` was the 4th certificate but is not in the `subjects` hashtable.

**Work environment**
 - Server OS: thehiveproject/cortex:3.0.1
- Cortex version: 3.0.1
 - Cortex Analyzer/Responder name: CIRCLPassiveSSL
- Cortex Analyzer/Responder version: 2.0

**Possible solutions**

1. Catch exception or test for hash presence in table
2. See server side what could be the issue if this is never expected to happen (ie. hash not in `subjects` hashtable)

PR #806, opened 2020-06-26
added check to validate that hash is in subject table

0xmilkmix:
Solution for:
- https://github.com/TheHive-Project/Cortex-Analyzers/issues/805

### DIFF
--- a/analyzers/CIRCLPassiveSSL/circl_passivessl.py
+++ b/analyzers/CIRCLPassiveSSL/circl_passivessl.py
@@ -34,10 +34,10 @@ class CIRCLPassiveSSLAnalyzer(Analyzer):
             certificates = []
         else:
             certificates = list(result.get(ip).get('certificates'))
+            subjects = result.get(ip).get('subjects', dict({}))
 
         newresult = {'ip': ip,
                      'certificates': []}
-        subjects = result.get(ip).get('subjects')
         for cert in certificates:
             if cert not in subjects:
                 continue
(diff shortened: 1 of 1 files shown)


garanews:
Thanks, 8[.]8[.]8[.]8 now works, but for example this 172[.]93[.]184[.]62 no.

garanews:
@0xmilkmix was you able to test with that ip?

jeromeleonard (approved):

jeromeleonard:
@0xmilkmix  @garanews, to fix issue mentionned by @garanews  maybe an option is to remove `subjects = result.get(ip).get('subjects')` of L40 and update the code L33 to:

```
        if not result.get(ip, None):
            certificates = []
        else:
            certificates = list(result.get(ip).get('certificates'))
            subjects = result.get(ip).get('subjects', dict({}))
```

0xmilkmix:
> @0xmilkmix @garanews, to fix issue mentionned by @garanews maybe an option is to remove `subjects = result.get(ip).get('subjects')` of L40 and update the code L33 to:
> 
> ```
>         if not result.get(ip, None):
>             certificates = []
>         else:
>             certificates = list(result.get(ip).get('certificates'))
>             subjects = result.get(ip).get('subjects', dict({}))
> ```

Hello @jeromeleonard , indeed that will work.

I implemented the fix that you suggested. Sorry for the delay !

jeromeleonard:
Thank you very much. We add it in the next release.

jeromeleonard:
@0xmilkmix  for your information, I added some code to extract specific observables from the report and also added some useful info in tags and descriptions. If you think other interesting info should be extract, feel free to share.  

image

PR #819, opened 2020-07-13
Update virustotal.py

TofBaasken:
Fix a bug in check_response. The code used to break when the variable results was of type list


dadokkio:
do you have any shareable example that is generating error?

TofBaasken:
Hi , basically all queries with get_report where there was a prior existing report available would crash the code , in that case the result variable (on line 47 on the current code) would be of type list instead of dictionary

TofBaasken:
btw this was tested on Cortex 3.0.0-1

PR #823, opened 2020-07-23
Updated VMRay Analyzer

53A-1:
This is a major rework of the VMRay Cortex Analyzer. It fixes a few outstanding issues with the existing implementation and adds support for a lot of features which allow a better control of the analysis process.
Thanks to TKCERT's @mback2k for his work on this.


mback2k:
If you don't mind, I could rebase the history a little bit to unsquash the first commit of this PR. I will provide you with the link to my branch once ready and then you could just force-push that to the branch of this PR.

mback2k:
@53A-1 my unsqashed branch with better commit messages is ready and can found here:

> https://github.com/TKCERT/Cortex-Analyzers/tree/develop-vmray-redux

From a content perspective it should be identical to your current develop branch used in this PR, at least a diff comes out empty.

You can perform the following steps to make your branch point to that HEAD and update this PR:

```
git remote add TKCERT https://github.com/TKCERT/Cortex-Analyzers.git
git checkout develop
git reset --hard TKCERT/develop-vmray-redux
git push -f <your-github-remote-name> develop
```

mback2k:
Any ETA on when this will be merged and part of a new release? @To-om @jeromeleonard

garanews:
Hi @mback2k in json I see tags required false, but if I do not set in config, when run I receive:
```
Error:
Traceback (most recent call last): File "/opt/Cortex-Analyzers/analyzers/VMRay/vmray.py", line 295, in <module> VMRayAnalyzer().run() File "/opt/Cortex-Analyzers/analyzers/VMRay/vmray.py", line 152, in run user_config=self.user_config, File "/opt/Cortex-Analyzers/analyzers/VMRay/vmrayclient.py", line 193, in submit_file_sample params["tags"] = ",".join(tags)TypeError: sequence item 0: expected str instance, NoneType found
```

Can you fix it?
Thanks!

53A-1:
@garanews Thank you for testing the implementation. Actually I think the problem here is a bug in the way Cortex handles empty lists, c.f. https://github.com/TheHive-Project/Cortex/issues/328
Nevertheless, I have implemented and pushed a Analyzer-centric workaround for this to this repo. Can you please test it again?

garanews:
It works, thanks!
Do you mind to add possibility in conf to switch the labels setting in something like short//extended ?
image

When have to manage tons of observables, I think it will be useful to have just a label for each analyzer and in this case the score will be the best one to keep.
Let me know your thoughts.

53A-1:
This could be implemented. However, since the score has been deprecated in the most recent version of VMRay anyway, I would suggest that this change is better suited for a version of the Cortex Analyzer that connects to VMRay 4.1 instead of the 3.3.1 patch discussed here. Whilst working on the result anyway I would also make the amount of labels configurable. Once this has been merged I will pull the new state, implement the changes based on that and create a new PR. Would this approach work?

garanews:
Will you add support for both 3.x and 4.x of vmray in same analyzer?
About deprecation of score, if any other field with a summary (such as safe/suspicious/malicious) will be present, we can use that one.

53A-1:
> Will you add support for both 3.x and 4.x of vmray in same analyzer?

Yes. Since up until now the API did retain the score in more recent versions too, it won't even be necessary to choose between the versions. Currently my idea is to include the new summary value along the existing ones. As for the filtered label set, I would include the new field if it is present in the results returned by VMRay, or the score if it is not.

> About deprecation of score, if any other field with a summary (such as safe/suspicious/malicious) will be present, we can use that one.

Yes, it works exactly as you describe.

garanews:
OK thanks, going to merge in dev right now.

PR #840, opened 2020-08-27
Added Shuffle workflow execution responder

frikky:
Could add more datatypes, but that's for later. This could e.g. execute something specific for every alert or case that is added. Could be good for specific enrichments etc.


garanews:
image
image
image
image

frikky:
Thanks @garanews! Will need to further customize this at some later point :+1:

Issue #847, opened 2020-09-08
[Bug] Mailer Responder is checking for recipient tags starting with "mail:"


jan4401:
**Describe the bug**
The Mailer Responder checks the tags list for the recipient mail address. 
In line 43 it checks for tags starting with "mail:" but with TheHive 4.0 the tags are using equal signs and not colons: "mail=" 
When using colons instead of equal signs in TheHive 4.0 it's automatically converted into an equal sign.

**To Reproduce**
1. Configure Mailer 1.0 Responder
2. Create Case
3. Create Tag: "mail:user@example.com"
4. Start Responder Mailer for this case.
5. Check Result of Responder for Error Message: recipient address not found in tags

**Expected behavior**
Normally the email tag should be detected by the responder python script.

**Complementary information**
Before clicking on the green hook:
![image](https://user-images.githubusercontent.com/49548458/92474648-7d916080-f1dc-11ea-93fb-4525ff3ac936.png)
After clicking:
![image](https://user-images.githubusercontent.com/49548458/92474709-96017b00-f1dc-11ea-83dd-a0cf04d1a82c.png)

**Work environment**

- Cortex version: 3.0.1-1
- TheHive version: 4.0.0-1
 - Cortex Analyzer/Responder name: Mailer
- Cortex Analyzer/Responder version: 1.0

**Possible solutions**
Change Line 43 from

`t.startswith("mail:")`

to

`t.startswith("mail=")`

**Additional context**
Could not test with TheHive 3.x.

jan4401:
see pull rquest https://github.com/TheHive-Project/Cortex-Analyzers/pull/848

viszsec:
any solutions to this so far?

dadokkio:
The pull proposed by @jan4401 is working. At the moment has been just merged in develop but you can test it.

PR #848, opened 2020-09-08
TheHive4: equal signs instead of colons for tags

jan4401:
TheHive4 does use equal signs instead of colons for tags
See issue: https://github.com/TheHive-Project/Cortex-Analyzers/issues/847


dadokkio:
Thanks, but probably is better if you check for both (mail: & mail=) in or to keep everything working also in thehive3.

jan4401:
@dadokkio okay thanks. I will try to add both

jan4401:
@dadokkio  i added an OR condition

jeromeleonard (approved):

PR #852, opened 2020-09-14
Velociraptor: Add upload functionality for flow results

weslambert:
- Adds the ability to upload flow results as a case observable.
- Other improvements

### DIFF
--- a/responders/Velociraptor/README.md
+++ b/responders/Velociraptor/README.md
@@ -6,7 +6,7 @@ The following options are required in the Velociraptor Responder configuration:
 
 - `velociraptor_client_config`: The path to the Velociraptor API client config.  
 (See the following for generating an API client config: https://www.velocidex.com/docs/user-interface/api/, and ensure the appropriate ACLs are granted to the API user).  
-- `velociraptor_artifact`: The name artifact you which to collect (as you would see it in the Velociraptor GUI)
-- `upload_flow_results`: Upload flow results to TheHive case (bool)
-- `thehive_url`: URL of your TheHive installation (e.g. 'http://127.0.0.1:9000')
-- `thehive_apikey`: TheHive API key used to add flow results/file(s) to a case
+- `velociraptor_artifact`: The name artifact you which to collect (as you would see it in the Velociraptor GUI).
+- `upload_flow_results`: Upload flow results to TheHive case (bool).
+- `thehive_url`: URL of your TheHive installation (e.g. 'http://127.0.0.1:9000').
+- `thehive_apikey`: TheHive API key used to add flow results/file(s) to a case.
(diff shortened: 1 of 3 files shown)


weslambert:
@dadokkio @garanews Any chance this could be included in the next release?

dadokkio:
I need @jeromeleonard feedback.. I always forget our policy regardind responder that uses thehive api :angel:

weslambert:
@jeromeleonard did you have any feedback?

weslambert:
Following up on this. Any thoughts @jeromeleonard @nadouani?

jeromeleonard:
hello @weslambert. The fact is that we do not recommend making a responder connect back to TheHive, unless there is no other choice. 

Like with Analyzers, Responders can add observables to a case when successfully run with the operation "AddArtifactToCase". For the moment, this cannot handle observables of type `file` but this is something we are considering in the coming releases of TheHive and Cortex. Seeing the growing need, we know this should be prioritized.  We are currently looking at the amount of work it needs to decide when it comes. I keep you posted.

weslambert:
Hi @jeromeleonard , thanks for the response.  

The use case is to have the results/artifacts sent back to TheHive for correlation with the case details, tracking, etc (in this case, as a zip file).  Additionally, some folks may have access to TheHive, but not to Velociraptor itself, so this may be the only way that they can view/retrieve this data alongside a case (approved artifacts/flows that are allowed to be run via responder, etc).

The current `VirustotalDownloader` responder here  essentially does the same/a very similar thing: 

https://github.com/TheHive-Project/Cortex-Analyzers/blob/master/responders/VirustotalDownloader/VirustotalDownloader.py#L52

weslambert:
@jeromeleonard Could it be possible to merge this for the time being, since this is not much different from the VT Downloader responder, then adjust as needed once the aforementioned functionality is available?

jeromeleonard:
Yes, we'll integrate it in the next release. Can I ask you to update the README.md file please ?  This will update to documentation page: https://thehive-project.github.io/Cortex-Analyzers/responders/Velociraptor/

weslambert:
Yep, will do.  Thanks!

weslambert:
@jeromeleonard @dadokkio would it be possible to add to this before merging?  I would like to make some updates.

dadokkio:
This has not been merged yes, so yes.
I wanted to test it, but then I had issues with velociraptor docker.

Regarding the necessity to use thehive api from responder I've for example added download sample functionality to [VirusTotal analyzer](https://github.com/TheHive-Project/Cortex-Analyzers/issues/899) without them, so there are still doubts of doing this.

nadouani:
@weslambert I'll merge this one, and you could update it afterward? Is this OK for you?

weslambert:
Hi @nadouani,

That would be fine, however I've not tested this in a while.

Issue #853, opened 2020-09-14
[FR] Update AnyRun Analyzer to include privacy setting


aacgood:
**Feature description**
When you run an analysis through AnyRun, there are 3 available privacy settings on the resulting analysis:
- `owner`: Only the user that submitted the analysis can view it
- `bylink`: Anyone with the link to the analysis can view (this is the default option of the API)
- `public`: Analysis is made public for all to consume

**Describe the solution you'd like**
Include the `opt_privacy_type` option with the URL and File submission requests

**Additional context**
While the default API option of `bylink` is going to be OK for the majority of the analysis cases, there is the potential for leakage or breaking of PAP if the link is further on-shared outside of your control.

PR #854, opened 2020-09-14
Included privacy type setting for URL/File analysis

aacgood:
Code to address FR:  https://github.com/TheHive-Project/Cortex-Analyzers/issues/853


garanews (changes requested):
Can you please keep version 1.0 this is not a big update and in this way no further templates and other actions are needed.

garanews (approved):
perfect

garanews:
Hi @aacgood with my tester license I was able to try private and bylink settings. both work like excpected.
I am sure that will work also public setting (but need hunter license)
image
image

PR #861, opened 2020-09-16
Fixes status_code 503 error caused by Cloudflare

omererdem:
- Added cloudscraper module and made requests with it to bypass Cloudflare protection when requesting details and location info.
- Updated requirements to use cloudscraper module.

### DIFF
--- a/analyzers/TalosReputation/TalosReputation.py
+++ b/analyzers/TalosReputation/TalosReputation.py
@@ -1,7 +1,7 @@
 #!/usr/bin/env python3
 # encoding: utf-8
 
-import requests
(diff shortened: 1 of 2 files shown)


dadokkio:
apart from the tab/spaces issue the pull works fine. If you can fix that I'll merge asap.

dadokkio (approved):

finchy:
Perhaps its because scraping of the data from talosintelligence.com is not allowed per the terms of service:

https://talosintelligence.com/reputation_center/support#terms

If someone wants to contact me regarding the use case and what you're trying to do here, we can chat.

PR #865, opened 2020-09-24
Small enhancement for encrypted zips

nielszcert:
We had a few cases that had .msg files with encrypted zips as an attachment. The FileInfo analyzer didn't handle that properly and it crashed.
To fix this I added a simple try except statement. I my test setup it analyzes the .msg and says in the attachment part that the zip is encrypted and cannot be read. I think this is better then not being able to analyze the whole email.


garanews:
added in https://github.com/TheHive-Project/Cortex-Analyzers/pull/915

PR #868, opened 2020-09-30
Add fqdn to Umbrella analyzer

LaZyDK:
Add fqdn to Umbrella analyzer


dadokkio:
Hi, just one quick question.. 
why are you mapping datatypes info to a dict? Probably is easier just to check:
`if self.data_type in ['fqdn', 'domain']`

LaZyDK:
You are absolutely right. I took this from another analyzer because this is pretty easy to expand in the future. 
We are using custom observable datatypes.

PR #872, opened 2020-10-03
Fixed some analyzers

emalderson:
- Fixed the JSON file of Maltiverse and Threatcrowd analyzers to remove unsupported observable types.
- Fixed URLhaus code, removing an if statement in which some non-existent fields of the JSON response are referenced in order to make the taxonomies building work

### DIFF
--- a/analyzers/Maltiverse/Maltiverse_Report.json
+++ b/analyzers/Maltiverse/Maltiverse_Report.json
@@ -5,7 +5,7 @@
   "url": "https://github.com/TheHive-Project/Cortex-Analyzers",
   "license": "AGPL-V3",
   "description": "Get the latest Maltiverse report for an hash, domain or an IP address.",
-  "dataTypeList": ["hash", "domain", "ip", "url"],
+  "dataTypeList": ["hash", "domain", "ip"],
   "command": "Maltiverse/maltiverse-client.py",
   "baseConfig": "Maltiverse",
   "config": {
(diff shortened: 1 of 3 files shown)


garanews (changes requested):
URL in Maltiverse  is fully supported:
https://app.swaggerhub.com/apis-docs/maltiverse/api/1.0.0-oas3 
https://github.com/maltiverse/python-maltiverse#24---url

FQDN in Threatcrowd  is fully supported (example: https://www.threatcrowd.org/searchApi/v2/domain/report/?domain=aoldaily.com and https://www.threatcrowd.org/searchApi/v2/domain/report/?domain=share.aoldaily.com )


md5_hash  and sha256_hash are valid fields: https://urlhaus-api.abuse.ch/

emalderson:
Thanks for your reply.
I know that those are supported on paper, but in fact using those types of observables with Maltiverse and Threatcrowd generates a lot of errors.
Furthermore, the taxonomies returned by URLHaus were always empty because those fields cannot be found in the json response. This has been found empirically analyzing the json response returned, not by reading the documentation.

garanews:
Please report some example.

emalderson:
Here are the outputs in Cortex after using Threatcrowd and Maltiverse with those types of observables:
https://i.imgur.com/I79KgvM.png
Here is the output in Cortex of URLHaus when something is found as malicious:
https://i.imgur.com/mZMOBfM.png

Issue #859, opened 2020-09-14
[OSCD Initiative] Develop Responder for Gmail


yugoslavskiy:
**Feature description**

Responder for Gmail that would be able to execute the following Response Actions:

- [RA4201](https://atc-project.github.io/atc-react/Response_Actions/RA_4201_delete_email_message/): Delete email message
- [RA3201](https://atc-project.github.io/atc-react/Response_Actions/RA_3201_block_domain_on_email/): Block domain on email
- [RA3202](https://atc-project.github.io/atc-react/Response_Actions/RA_3202_block_sender_on_email/): Block sender on email
- [RA5201](https://atc-project.github.io/atc-react/Response_Actions/RA_5201_unblock_domain_on_email/): Unblock domain on email
- [RA5202](https://atc-project.github.io/atc-react/Response_Actions/RA_5202_unblock_sender_on_email/): Unblock sender on email

**Describe the solution you'd like**

It could be done via Gmail API using the following methods:

- Delete email message — https://developers.google.com/gmail/api/v1/reference/users/messages/delete
- Block domain on email — https://developers.google.com/gmail/api/guides/filter_settings (automatic redirection to trash)
- Block sender on email — https://developers.google.com/gmail/api/guides/filter_settings (automatic redirection to trash)
- Unblock domain on email — https://developers.google.com/gmail/api/guides/filter_settings (disabling filter)
- Unblock sender on email — https://developers.google.com/gmail/api/guides/filter_settings (disabling filter)

strassi:
I would implement this responder. @yugoslavskiy  do you have some hints if the API is the same for G Suite and private GMail?

strassi:
OK. If I understood correctly everything is based on the OAuth 2.0 Authentication / Authorization. So the API is the same for GSuite/Google Workplace and a private Gmail Account.

According to 
 - https://support.google.com/a/answer/7281227?hl=en&ref_topic=10021546
 - https://support.google.com/a/answer/162106?hl=en&ref_topic=10021546

Apps can be given a _service account_ which authenticates via OAuth 2.0. This _service account_ is able to connect to G Suite / Google Workspace data without individual user consent. Otherwise users must authorize the app access to their data.

OAuth 2.0 authorizes actions via scopes. The valid Gmail scopes can be under https://developers.google.com/identity/protocols/oauth2/scopes#gmail. For the feature we might need the following scopes:

| Action | Scope |
|:------:|:--------|
[...]

PR #891, opened 2020-11-06
[OSCD Initiative] add Gmail responder

strassi:
Implementation of the Gmail Responder for the OSCD Initiative. Can only be used with a valid Gsuite Account.

This fixes Issue #859


responders/Gmail/Gmail.py
```diff
@@ -0,0 +1,224 @@
...
+    def trash_message(self, case_id, query):
+        """Moves specified message into trash. this emails can be recovered if false-positive
+        """
+        gmail_observables = self.__get_gmail_subjects(case_id, And(Eq("dataType", "mail"), EndsWith("data", self.__gmail_domain)))
+        for observable in gmail_observables:
+            resource = self.gmail_impersonate(observable["data"])
+            try:
+                response = resource.users().messages().list(userId=observable["data"], q=query).execute()
+                for message in response.get("messages", []):
+                    result = resource.users().messages().trash(userId=observable["data"], id=message["id"]).execute()
```
yugoslavskiy, line 107: Hello @strassi! 
Great job with the responder!
What do you think about adding `.delete` function as well? 
There are some cases when the email's nature is so dangerous, so moving to trash is not an option (as the users still could read it if they go and check the trash), only complete deleting could be used.

responders/Gmail/Gmail_UnblockSender.json
```diff
@@ -0,0 +1,76 @@
+{
+    "name": "Gmail_UnblockSender",
+    "version": "1.0",
+    "author": "David Strassegger, OSCD",
```
yugoslavskiy, line 4: Would you mind extending it to the 'oscd.community' or '@oscd_initiative'? 
It would be more clear and help to engage more people to the next sprints (:

strassi: Yeah no problem. I changed it to `@oscd_initiative`

responders/Gmail/Gmail.py
strassi: > What do you think about adding .delete function as well?

I thought about adding a delete service as well, but did not implement it because of these reasons:

1. The difference between the actions `.trash` and `.delete` might not be clear at first. A False-Positive could lead to data loss if `.delete` was used.
2. Deletion is done by running a Gmail search query against the users mailbox. A faulty query could erase an entire mailbox.

In my opinion, the `.trash`ing of messages is a safe default that enables easy recovery. I also might argue, that a user does not dig up random trashed messages. In my experience they search for specific trashed messages in case of recovery.

 > There are some cases when the email's nature is so dangerous, so moving to trash is not an option (as the users still could read it if they go and check the trash), only complete deleting could be used.

Can you provide an example for such a message?

yugoslavskiy: I see your point, and it totally makes sense. This topic is quite complex and has a few not obvious things to consider. Let me highlight some of them:

> The difference between the actions `.trash` and `.delete` might not be clear at first. A False-Positive could lead to data loss if `.delete` was used.

Removing the email is a Response Action from the Eradication stage:

image

This stage is harmful by definition — it's about deletion files, mails, services, and any kind of threat actor presence.
At the same time, it's not a stage when you can change your mind and re-classify an Incident to a False Positive. 
That a task for the Identification stage:

image

The Identification stage is also a stage when you collect the evidence, make a copy of it, and put it in a safe place for further analysis/processing/legal procedures. 
Getting to the Eradication stage not having a copy of evidence is not the best way to handle the incident. 

> Deletion is done by running a Gmail search query against the users mailbox. A faulty query could erase an entire mailbox.

I believe, the nature of the Eradication stage could be extremely harmful in the wrong hands. But IR people supposed to know what they are doing, configuring firewall rules, deleting files, changing configuration, etc. I believe that it is a normal thing, that some Responder during the Eradication *could* cause issues like the one you've described. But I don't think that this is somehow connected with your Responder, as this is a tool. It should be powerful. If somebody erased user's mailbox entirely, it's totally not the Cortex Responder issues. It's the Incident Response specialist issue.

> Can you provide an example for such a message?

Yep. There were many cases with malicious macros in MS Office files, attached to the emails.
User opens it, does a few clicks — boom, the second stage is downloaded, executed, and everything is on fire. 

The most horrible part of it is that some users in the organization "A" (which we were monitoring) were sending this email to their friends (asking "hey do you know why it doesn't work? Could you please try to open it for me?" etc) in the organization "B", that fortunately was also under our monitoring. 
So one incident related to one organization became a multi-organization incident. 

We also need to keep in mind that threat actors can ask victims to check their trash/spam mailboxes using other channels (LinkedIn, instant messengers, other social media). 
That's why in some cases we have to completely remove the malicious emails.

---

I believe that having both `.trash`ing and `.delete`ing options is the best way (: 
We can develop special notification about `.delete` function, providing all the details and warnings to the users of your Responder, this way, providing people with the freedom to choose. 

What do you think, @strassi?

strassi: I thought about your arguments concerning the IR eradication stage. After reassessing my thoughts about the `.delete` API call I came to following conclusion.

I would swap the `.trash` with the `.delete` API.
This way the action cannot be misunderstood and it serves the single purpose of cleaning out a mailbox as a IR specialist would expect. A second advantage of simply swapping the API calls would be: Less code, less errors.

Unfortunately I cannot delete messages via the [filter settings API](https://developers.google.com/gmail/api/guides/filter_settings). Filters can only move the message to trash at best. 

Would you see this is as a possible solution?

yugoslavskiy: I've found one, I am not sure if that's the best of possible options, but it works. 

The workflow is the following:

1. Grab all users in the domain using Admin API
2. Execute filter query per each account 
3. Delete email by message ID if found

How to make it work:

1. Enable [Admin API](https://developers.google.com/admin-sdk)
2. Add the `https://www.googleapis.com/auth/admin.directory.user.readonly` to the existing service account scopes
3. Request new `credentials.json` (I am not sure if this step is required, did it just in case as the new permissions could not be applied to the old credentials. At least that's how it works for other APIs I've worked with)
4. Use the code snippet below to delete messages using filter query:

```python
#!/usr/bin/env python3

from google.oauth2 import service_account
from googleapiclient.discovery import build

SUBJECT = "impersonation@domain.com"

GMAIL_SCOPES = [ 
    "https://mail.google.com/", 
    "https://www.googleapis.com/auth/gmail.settings.basic",
]

ADMIN_SCOPES = [ 
    "https://www.googleapis.com/auth/admin.directory.user.readonly"
]

SERVICE_ACCOUNT_FILE = "/path/to/credentials.json"

TARGET_DOMAIN='domain.com'  # Service account will access data from this domain.

#
# Fill in a list of user IDs (emails)
#

credentials = service_account.Credentials.from_service_account_file(
        SERVICE_ACCOUNT_FILE, subject=SUBJECT, scopes=ADMIN_SCOPES)

admin_service = build("admin", "directory_v1", credentials=credentials)
results = admin_service.users().list(domain=TARGET_DOMAIN).execute()
users = results.get('users', [])


all_users = []
for user in users:
    all_users.append(user['primaryEmail'])

#
# Remove emails that match our query for each user
#

query = "from:badguy@malicious.com"

for user_id in all_users:
    credentials = service_account.Credentials.from_service_account_file(
        SERVICE_ACCOUNT_FILE, subject=user_id, scopes=GMAIL_SCOPES)
    gmail_service = build("gmail", "v1", credentials=credentials)
    response = gmail_service.users().messages().list(userId=user_id,q=query).execute()
    if response.get('messages'):
        for message in response['messages']:
            gmail_service.users().messages().delete(userId=user_id, id=message['id']).execute()

```

Let me know what do you think about it (:

strassi: In my opinion the responder should not be run over the entire user base on its own. The observables in a case serve the purpose of documenting IOCs and affected objects/subjects. At least this is what I use them for.

I think it is quite a good practice to isolate the work of a responder does as much as possible. Too much logic takes away the control of the incident responder.

I would propose changing the `.trash` call to `.delete`, but leave the filter behavior as it is for now. If a IR specialist demands a "erase" functionality over the whole user domain we can add it in later via a new pull request. 

You could think of this responder with the current feature set as a MVP.

yugoslavskiy: If you would like to finish at this point, that's totally fine (: You did a great job, developing the Responder.

I will create an issue and provide my snippet that could help to develop the following functionality:

- [RA4201](https://atc-project.github.io/atc-react/Response_Actions/RA_4201_delete_email_message/): Delete email message from all the domain's users
- [RA2201](https://atc-project.github.io/atc-react/Response_Actions/RA_2201_list_users_opened_email_message/): List users opened email message
- [RA2203](https://atc-project.github.io/atc-react/Response_Actions/RA_2203_list_email_message_receivers/): List email message receivers

I think that your responder will have a great future and a lot of users will benefit from it!

strassi: Yes, I would propose to finish the responder.

I'm going to swap the `.trash` with the `.delete` api call.

yugoslavskiy:
Hello @strassi!

Sorry for the delay, it was a quite busy time.

I've tested your responder and got a few comments:

1. Please replace `urllib` with `urllib3` in `requrements.txt`, as it doesn't work the way it is
2. It seems that if a non-company email (sender) was blocked, it's not getting a proper tag, which is required for further unblocking. So we can block a sender, but cannot unblock. Could you please fix it? I would be able to proceed to test your Responder after that

strassi:
Hi @yugoslavskiy!

1. not a problem
2. I just had a quick look at the code but can't spot issues that would match your problem description. The observables being blocked (sender) does not get a tag. The tags are set on the gmail addresses of the case. This is done because filters are set in the context of the gmail user.

Expected behavior (that worked in my tests, in November):

1. Add `name.surname@gmail.com` observable to case.
2. Add `hacker@evil.com` observable to case.
3. Start Gmail_BlockSender Responder on `hacker@evil.com`.
4. `name.surname@gmail.com` gets tag `gmail_filter:hacker@evil.com:<gmail_filter_id>`
5. Start Gmail_UnblockSender Responder on `hacker@evil.com`.
6. Tag on `name.surname@gmail.com` should get removed.

Can you provide a some sort of screenshots or output along the way of the described test run?

yugoslavskiy:
Hi @strassi! 

Here is how it was:

1. Add yugoslavskiy@gmail.com observable to a case.
2. Add daniil@yugoslavskiy.com observable to a case.
3. Start Gmail_BlockSender Responder on daniil@yugoslavskiy.com.
4. yugoslavskiy@gmail.com doesn't gets tag gmail_filter:daniil@yugoslavskiy.com:<gmail_filter_id>
5. Start Gmail_UnblockSender Responder on daniil@yugoslavskiy.com.
6. It doesn't start with an error: "No valid filter tag found on observable. Tags: ['<random tag that I had to create while adding the observable>']"

Here is the screenshots:

image

image

strassi:
I took the TheHive 3.4.0 Training VM for debugging the problem. Additionally I disabled the Gmail API calls in `block_messages` and `unblock_messages`  to test the tagging mechanism for the thehive. The tagging seems to work:
image

image

To get this work in the training VM I had to upgrade `thehive4py` pip3 packages to 1.8.1. The responder uses the `thehive4py.query` API. Without the upgrade the responder fails because `EndsWith` is not defined.

@yugoslavskiy can you check if the installed version of your `thehive4py` is >= 1.8.1. If not please upgrade the thehive4py version and repeat your tests. 

I'm going to set the requirements.txt to include the minimum version for the library to exclude future problems of this kind.

yugoslavskiy:
Hello @strassi! 

Hope you are doing well (: 

I had `thehive4py` version 1.8.1 at my previous tests:

```
$ python3 -m pip show thehive4py | grep 'Version'
Version: 1.8.1
```

So I decided to update all the packages to the versions you fixed in your [last commit](https://github.com/TheHive-Project/Cortex-Analyzers/pull/891/commits/0c0525165f137aad5f9bfba93f34a94fe150557d), and here is what I got:

```
ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts.
docker-compose 1.24.1 requires jsonschema<3,>=2.5.1, but you have jsonschema 3.2.0 which is incompatible.
docker-compose 1.24.1 requires requests!=2.11.0,!=2.12.2,!=2.18.0,<2.21,>=2.6.1, but you have requests 2.25.0 which is incompatible.
botocore 1.14.17 requires urllib3<1.26,>=1.20; python_version != "3.4", but you have urllib3 1.26.3 which is incompatible.
```

It seems that it's time to move to the docker, but I have not idea how to make it work there, even tho I see the Dockerfile you've created.
OK let's hope that problem was somewhere on my side, or in my workflow. 
I will repeat the steps anyway and let you know how it went.

yugoslavskiy:
Here is how it was:

1. Create a new case "Gmail Responder Test Case"
2. Add `yugoslavskiy@cindicator.com` (it utilizes Gmail) observable to a case.
3. Add `daniil@yugoslavskiy.com` observable to a case.

Here is how it looked at that step:

image

4. Execute `Gmail_BlockSender` Responder on `daniil@yugoslavskiy.com`. Here is the result:

Here is the info about successful execution:

image

Here is the list of observables (no tags created):

image


So it seems that the result is the same, `yugoslavskiy@cindicator.com` doesn't gets tag `gmail_filter:daniil@yugoslavskiy.com:<gmail_filter_id>`.

Could you please give me a hint on the debugging? It seems that you somehow spot the error with the old version of `thehive4py`. Maybe I will see a different error in my environment using your approach.

strassi:
Hi @yugoslavskiy !

I'm doing fine besides wondering about this bug 🤔 

So far we see that

-  the responder results in a success state (so no exceptions)
-  the blocked observable gets a tag (so tagging works fine)
-  your `thehive4py` version is correct (so `EndsWith()` from `thehive4py.query` is available)

I'm wondering if the responder gets any gmail obervables from the case. You could change line 145
```
145        self.report({'message': "Added filters"})
```
to 
```
145        self.report({'message': "Added filters to {}".format(gmail_observables)})
```

this should print the list of gmail observables to the responders output. Furthermore you could check if the filters are set in the Gmail settings of the respective gmail address (e.g. `yugoslavskiy@cindicator.com`).

yugoslavskiy:
Yeah, the filters are set, that's for sure. I had to delete the previous one manually. 
Also, removing filters via the `Gmail_UnblockSender` responder works fine if creating the proper tag manually.

The only issue with the `Gmail_UnblockSender` is that it breaks if there are two fiter tags present. The first one — old, that has been removed, the second one — an active filter:

image

So it breaks on first and didn't try to proceed with the second one. Maybe the main issue here is that it doesn't remove the filter tag from the observable after deleting the real filter in the Gmail (after successful execution), which (IMHO) it should do.

I've updated the line you've suggested, and here is what I got:

```
{
  "message": "Added filters to [{'updatedAt': 1614729931036, 'tlp': 0, 'tags': ['test', 'gmail_filter:daniil@yugoslavskiy.com:ANe1BmiYy6Kn74pGfhlfV1nDOUZmAyfjASxB3Q'], 'data': 'yugoslavskiy@cindicator.com', 'ioc': False, 'reports': {}, 'status': 'Ok', 'sighted': False, 'createdAt': 1614712731513, 'createdBy': 'admin', 'message': 'test', 'dataType': 'mail', 'updatedBy': 'admin', 'startDate': 1614712731514, '_type': 'case_artifact', '_routing': 'sZxf9HcBfFDkyb-x_nuK', '_parent': 'sZxf9HcBfFDkyb-x_nuK', '_id': 'bfc80b63e6be12a61ddbb40b997163ee', '_seqNo': 121, '_primaryTerm': 2, 'id': 'bfc80b63e6be12a61ddbb40b997163ee'}]"
}
```

It seems that it wants to add a new tag or consider that it has been added, but it haven't:

image

Also, if we consider this one as a UI bug, and try to execute `Gmail_UnblockSender`, we will see the same error:

image

So the problem is that Cortex for some reason cannot add the tag, and does it silently. At the same time it somehow added the "gmail:handled" tag.

So I was about to ask what permissions did you provide cortex user account with, but decided to add `write` permissions to it and check if it will start working (it had only `read` permission before).

So it did. Now it creates tags, removes tags, does its job, works fine. 
Tomorrow will test the rest of the responders from your PR. 

Thank you for your work and for your patience, man!

strassi:
> So I was about to ask what permissions did you provide cortex user account with, but decided to add write permissions to it and check if it will start working (it had only read permission before).

I see the problem. For my tests I used the API key of the thehive _admin_ user in the training VM. This user does have write permissions.

I just realized, that the API provides a `get_current_user` [function](https://github.com/TheHive-Project/TheHive4py/blob/6bd75773889dac6c664310b187d0b264607075cf/thehive4py/api.py#L144), which returns a user object. Maybe the user object does have a property _roles_. This could be checked for _read,write_. The responder could fail if it does not have sufficient user rights. This might fix these kind of problems in the future.

Nice analysis  @yugoslavskiy; would not have thought of the api key permissions

yugoslavskiy:
Hello @strassi! 

I've tested the other functions and everything works smoothly. 
I believe we can summon maintainers and ask them for merge (: 

@nadouani @jeromeleonard guys, is there anything else required from our side to proceed with the merging?

yugoslavskiy:
Hello @nadouani @jeromeleonard!

Is there anything required from our side to proceed with merging? (:

yugoslavskiy:
Hello @strassi! Could you please change the base (target branch) to the `feature/oscd`? 
We would like to have 1 final PR with the entire OSCD sprint work.

PR #906, opened 2020-11-22
[OSCD Initiative] Add Azure Authentication Token Revokation Responder

Dmweiner:
Hello, please see the supplied AzureTokenRevoker responder .py, .json, and requirements.txt files.

Two outstanding issues within the responder's .py codew:
1.) I was not able to clear up, is what the get_params data path was for a submitted UPN within a case. Would it be "data.data"? Or something else?

2.) Will a self.report statement with a message be sufficient as a return of success? Or is there some other object I should be passing back to TheHive?

### DIFF
--- a/analyzers/URLhaus/URLhaus_analyzer.py
+++ b/analyzers/URLhaus/URLhaus_analyzer.py
@@ -35,7 +35,8 @@ class URLhausAnalyzer(Analyzer):
         namespace = "URLhaus"
 
         if raw['query_status'] == 'no_results' \
-        or raw['query_status'] == 'ok' and raw['md5_hash'] == None and raw['sha256_hash'] == None:
+        or (raw['query_status'] == 'ok' and not raw.get('md5_hash', None) \
+        and not raw.get('sha256_hash', None)):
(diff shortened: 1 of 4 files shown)


yugoslavskiy:
Hello @Dmweiner! Sorry for the delayed reply. We still work on the OSCD sprint finalization. 
Let's try to tag the maintainers and get your questions answered! 

@nadouani @jeromeleonard guys could you please give us a hand?

nadouani:
> Hello, please see the supplied AzureTokenRevoker responder .py, .json, and requirements.txt files.
> 
> Two outstanding issues within the responder's .py codew:
> 1.) I was not able to clear up, is what the get_params data path was for a submitted UPN within a case. Would it be "data.data"? Or something else?
> 
> 2.) Will a self.report statement with a message be sufficient as a return of success? Or is there some other object I should be passing back to TheHive?

Hello @Dmweiner 

1/ using `data.data` you get access to the case send by TheHive to the responder. Where do you store the UPN within a case?
2/ technically this is enough, I would return an operation like adding a tag to the case or observable.

PR #907, opened 2020-11-23
fix template for floss

dadokkio:
fix template issue


dadokkio:
Can be closed since integrate in #915

PR #911, opened 2020-12-09
GreyNoise - Update API to new version 3

markus-nclose:


markus-nclose:
https://github.com/TheHive-Project/Cortex-Analyzers/issues/912

analyzers/GreyNoise/greynoisev3.py
```diff
@@ -0,0 +1,132 @@
...
+class GreyNoiseAnalyzer(Analyzer):
+    """
+    GreyNoise API docs: https://developer.greynoise.io/reference#noisecontextip-1
+    """
+
+    def run(self):
+
+        if self.data_type == "ip":
+            api_key = self.get_param('config.key', None)
+            api_client = GreyNoise(api_key=api_key, timeout=30)
```
bradchiappetta, line 20: @markus-nclose : can we also get an integration name passed here?

Example:
api_client = GreyNoise(api_key=api_key, integration_name="greynoise-cortex-analyzer-v3.0")

We're working to set this up as our standard going forward.

danielbrowne: Added as requested above, looks like the only use is in the user-agent header?

bradchiappetta: Thanks, yes that is correct, this is essentially a user-agent string for us to see on the backend.  It helps us understand the adoption and use of integrations with our service and we can also proactively let users know there is a new version available if we see high usage of older versions.

dadokkio:
I merged the pull in develop because I really liked the api approach and then I found out that I was working on it in different branch to fix documentation so I had to merge our forces :smile: 

You can find updates here: 
- code: https://github.com/TheHive-Project/Cortex-Analyzers/tree/feature/update_docs/analyzers/GreyNoise
- template: https://github.com/TheHive-Project/Cortex-Analyzers/blob/feature/update_docs/thehive-templates/GreyNoise_3_0/long.html

Changes:
- removed requests since it's not used anymore
- used python3 in #! 
- moved templates in _3_0 folder and update to new json results

Sorry for the confusion! If you have any other feedback (eg. how to improve docs) let me know!

Issue #839, opened 2020-08-14
[Improvement] FileInfo should include actual attachments in the report


xeaon:
### Request Type
Feature / Improvement

### Description
Since extracting file observables from reports was implemented in TheHive-Project/TheHive#982 (i still can't wrap my head around how I'm supposed build the report that TheHive picks the file due to lack of knowledge and documentation) it would be extremely useful, if the FileInfo Analyzer would this, especially in the Outlook submodule. 

### Example workflow after Improvement
1. upload any malspam.msg which includes a malicious .doc and analyze with FileInfo
2. import malicious .doc attachment provided in the report as file observable
3. scan attachment to gain more insight real quick

If you could point me to the right direction, for example how a report has to look like to trigger TheHive to pick up file observables, I'd try to implement this improvement by myself and create a pull request.

dadokkio:
Some wip [not yet fully functional] from sample email:
![image](https://user-images.githubusercontent.com/355183/99991563-7cbca380-2db5-11eb-82ed-e31d88cdf399.png)

xeaon:
Thank you for your efforts so far! Let me know if I can help in any way.

v1p3r0u5:
@dadokkio Exactly what I'm looking for. Any chances that this will also be merged into TH4 soon?
Many thanks for your effort. Highly appreciated.

xeaon:
> @xeaon merged in devel with #915
> It needs latest cortexutils v 2.1.0

Thank you and sry for the late response. I'll give it a try!

PR #314, opened 2018-07-27
New analyser: File Strings

j-i-m-b-o:
A simple analyser to find printable strings within files. 

I've personally found this analyser to be a useful first step when dealing with various types of unknown file so wanted to make it available to others.

PR #915, opened 2020-12-14
Update Fileinfo

dadokkio:
Some new feature/issues:

#839 (only for TH3 at the moment). Each submodule can return also an observables list that can be then imported.
At the moment has been added to submodule_outlook but similar approach can be used on each module.

#314 .. following 3c7 work adds strings, floss and rankstrings as submodule.


garanews:
image
image
image

Issue #920, opened 2020-12-25
[FR] Responder which sends a mail with a detailed incident status 


mkcorpc:
**Feature description**
Mail a detailed status information of an incident case including incident id, name, severity, custom fields, etc. The mail is sent to recipients specified by tags prefixed with 'mail='. The responder respects tlp definitions. For tlp:amber mail address and for tlp:green mail domains must be pre-defined in the configuration. For tlp:red sending mails is denied. The responser also uses thehive4py to collect information about the status of the tasks of the incidents.

**Describe the solution you'd like**
A new responder should be developed. A second options would be to replace the Mailer responder. I am not sure, but i think the 'Mailer' just sends an empty mail.

**Additional context**
No additional context

dadokkio:
Closing issue since pull has been merged

PR #921, opened 2020-12-25
Feature/mail incident status

mkcorpc:
Please find enclosed the implemented responder.

[FR] Responder which sends a mail with a detailed incident status #920

### DIFF
--- a/responders/MailIncidentStatus/MailIncidentStatus.json
+++ b/responders/MailIncidentStatus/MailIncidentStatus.json
@@ -1,0 +1,98 @@
+{
+  "name": "MailIncidentStatus",
+  "version": "1.0",
+  "author": "Manuel Krucker",
+  "url": "https://github.com/TheHive-Project/Cortex-Analyzers",
+  "license": "AGPL-V3",
+  "description": "Mail a detailed status information of an incident case. The mail is sent to recipients specified by tags prefixed with 'mail='. The responder respects tlp definitions. For tlp:amber mail addresse and for tlp:green mail domains must be pre-defined in the configuration. For tlp:red sending mails is denied. The responser also uses thehive4py to collect information about the status of the tasks of the incidents.",
+  "dataTypeList": ["thehive:case"],
+  "command": "MailIncidentStatus/mailincidentstatus.py",
+  "baseConfig": "MailIncidentStatus", 
+  "configurationItems": [
+    {
+      "name": "from",
+      "description": "email address from which the mail is send",
+      "type": "string",
+      "multi": false,
+      "required": true
+    },
+    {
+      "name": "smtp_host",
+      "description": "SMTP server used to send mail",
+      "type": "string",
+      "multi": false,
+      "required": true,
+      "defaultValue": "localhost"
+    },
+    {
+      "name": "smtp_port",
+      "description": "SMTP server port",
+      "type": "number",
+      "multi": false,
+      "required": true,
+      "defaultValue": 25
+    },
+    {
+      "name": "smtp_user",
+      "description": "SMTP server user",
+      "type": "string",
+      "multi": false,
+      "required": false,
+      "defaultValue": "user"
+    },
+    {
+      "name": "smtp_pwd",
+      "description": "SMTP server password",
+      "type": "string",
+      "multi": false,
+      "required": false,
+      "defaultValue": "pwd"
+    },
+    {
+      "name": "mail_subject_prefix",
+      "description": "Prefix of the mail subject",
+      "type": "string",
+      "multi": false,
+      "required": false,
+      "defaultValue": "Incident Case Notification: "
+    },
+    {
+      "name": "mail_html_style_tag_content",
+      "description": "The css content of the style tag for the HTML mail body. Define table, th, hd, .first, and .second elements.",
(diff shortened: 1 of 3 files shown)


responders/MailIncidentStatus/MailIncidentStatus.json
```diff
@@ -0,0 +1,98 @@
...
+      "required": false,
+      "defaultValue": "table { border: 1px solid black; border-collapse: collapse; text-align: left; vertical-align: top; th { border: 1px solid black; border-collapse: collapse; text-align: left;} td { border: 1px solid black; border-collapse: collapse; text-align: left;} .first { width: 150px; min-width: 150px; max-width: 150px; background-color: #ffe8d4; } .second { background-color: #d7d9f2;}"
+    },
+    {
+      "name": "tlp_amber_mail_addresses",
+      "description": "Mail addresses which are allowed to receive tlp:amber classified incidents",
+      "type": "string",
+      "multi": true,
+      "required": false,
+      "defaultValue": "Incident Case Notification: "
```
ntcong, line 73: this should be ["Incident Case Notification: "] or better ["address@example.com"]

mkcorpc: Yes correct. 

Sorry it is my first pull request. Do I need to do something to confirm this change?

ntcong: I don't have reviewer status so it's just a comment

I tried your plugin on thehive4 and it displays a list of separate character instead of a string

mkcorpc: > I tried your plugin on thehive4 and it displays a list of separate character instead of a string

I dont know what your a meaning with 'separate characters'. Do you speak about an configuration option?

The following mail is sent by the plugin:
mail

ntcong: image

it looks like this from the settings page

dadokkio:
I'm going to test the responder now. 
Quick suggestion, can you remove ssl, email and json from requirements. Those are standards libs and trying to install requirements with pip it tries to install wrong ssl lib instead.

responders/MailIncidentStatus/MailIncidentStatus.json
dadokkio: I can confirm ntcong error and suggestion.
First line is with his fix `"defaultValue": ["Incident Case Notification:"]`
second whith standard code "`defaultValue": "Incident Case Notification: "`
image

dadokkio:
Some issues during tests:

- My case don't have data.updatedAt value set so it fails when tries to divide it by 1e3
- I've seen that you have commented out server.send_message(msg) .. it should be ok since you put From and To in the msg so are not required as parameters. Also [mail_address] returns an error since a list is not a valid option
- I've custom fields that are not set in this case, but the create_HTMLTable function returns error because i[1] is then None and cannot be concatenate
- When you check for email in tags can you please check for both "mail=" and "mail:"? TH3 and TH4 have different behavior on tags management and you need to check both.

mkcorpc:
I have implemented the following changes:
- [X] Removed ssl, email and json from requirements
- [X] Removed the comment 'server.send_message(msg)'
- [ ] [mail_address] returns an error since a list is not a valid option
- [ ]

mkcorpc:
Addjusted json configuration. 
- [X] Deleted default value for tlp_amber_mail_addresses
- [X] Deleted default value for tlp_green_mail_domains

dadokkio:
I was able to run the plugin :smile:, I just put small comments in the code.
The multilined ().format syntax for text I proposed for the summary could be also used in email creation.

mkcorpc:
Sorry, but I don't know where i can find your comments. How can I see these?

responders/MailIncidentStatus/mailincidentstatus.py
```diff
@@ -0,0 +1,272 @@
...
+        html += "<meta charset=\"UTF-8\"/><meta http-equiv=\"Content-Type\" content=\"text/html\"; charset=\"utf-8\"/>\n"
+        # styles
+        html += "<style>" + self.get_param('config.mail_html_style_tag_content',None,"Error loading config 'config.mail_html_style_tag_content'") + "</style>\n"
+        #html += self.get_HTMLStyle()
+        html += "</head>\n\n"
+        html += "<body>\n"
+        html += "<table width=\"100%\" cellpadding=\"0\" cellspacing=\"0\" border=\"0\" bgcolor=\"#FFFFFF\" align=\"center\">\n"
+        html += "<colgroup><col class=\"first\"/><col class=\"second\"/></colgroup>\n"
+        html += "<tr><td colspan=\"2\">Incident Status Report</td></tr>\n"
+        for i in two_tuple_list:
```
ntcong, line 226: updatedBy can be None: ('updatedBy', None) which will trigger error:
`html += i[1] TypeError: can only concatenate str (not "NoneType") to str`

dadokkio:
I updated your code with my and @ntcong suggestions. Let me know your opinion :+1: 
It seems all changed cause I've black formatting on save :smile:

mkcorpc:
I have "accepted" all conversations.

PR #923, opened 2021-01-09
Add IVRE Analyzer

### LINKED ISSUE #922: [IVRE](https://ivre.rocks/)-based analyzer
### Request Type
Analyzer

### Description
I'm writing an [IVRE](https://ivre.rocks/)-based analyzer for my needs, and I'd like to publish it to this repository.

IVRE is an open-source network recon framework (see the [repository](https://github.com/cea-sec/ivre)), that stores intelligence about hosts (and flows but only hosts are, for now, concerned by the plugin) collected either passively (using Zeek) or actively (using Nmap, Masscan, ZGrab2, ZDns, and soon [Nuclei](https://github.com/cea-sec/ivre/pull/1068).

The plugin will probably be similar to the [IVRE plugin in Yeti](https://github.com/yeti-platform/yeti/tree/master/contrib/analytics/ivre_api).

Please let me know if you have any questions, suggestions or ideas.

p-l-:
Fixes #922.


dadokkio:
It's possible to add mongodb ip as config in order to be able to use a separate docker instance of ivre?
Will you add templates and screenshots for long and short templates?

p-l-:
Yes I want to add templates, that's just a "work in progress" PR so that others can have a look.
You normally configure IVRE using its config file (system-wide: `/etc/ivre.conf`, or user-specific: `~/.ivre.conf`).

dadokkio:
It was the first time yesterday I've installed and used IVRE so I didn't know it [Probably a readme with some indication could be also useful :angel:]

The "issue" is that cortex in my case is on a machine, while ivre is dockerized on another one. 
So cortex doesn't know where the db is and I think is not able to read that conf too. Or I'm missing something?

p-l-:
I am currently in the process of improving the `DBHttp` backend so that one IVRE instance can access data through the another IVRE instance's HTTP server (using a configuration like `DB = "http://192.168.0.1/cgi/"`. In the meantime, you can expose the MongoDB port of the `ivredb` container (usually 27017).

dadokkio:
Ok, it takes me time to understand properly your previous message.
I've added "/etc/ivre.conf" on cortex machine and now it's looking for db in the right place :+1: 
In any case, I think that having a file based conf is not so user-friendly when your run dockerized analyzer.

p-l-:
@dadokkio This should work as you'd expect now. What do you think?

dadokkio:
Yes, that should permit to use also the dockerized version of analyzer because in that case you don't have external files. The image is built, executed and then deleted as is. 
I'm going to test it again today, I'm just adding a missing } in json to test it :+1:

dadokkio:
I was able to test also templates and they works fine.

image


Are you still in wip or can we consider this pull ok? 
The only suggestion I can provide it's to create a README with some indication on how to start and configure everything.

p-l-:
Thanks a lot @dadokkio for your help. I think this PR is now ready for review, I hope I did not miss something else.

p-l-:
BTW I think the "status:needs-template" tag no longer applies.

p-l-:
Thanks @dadokkio!

PR #950, opened 2021-03-01
update full report link for VirusTotal GetReport Template

buzzdeee:
similar to VirusTotal_Scan, use the content.permalink to link back to the full report


dadokkio:
Second ng-if is not necessary, but the pull is ok :+1:

PR #962, opened 2021-03-25
Fix splitting inside parsed header extraction.

ch0wm3in:
I dont know what initially caused this to be(besides the comment about body being parsed also, but my testing with a few different eml files i cannot replicate it original dev was probably using an old version of it, and future commits continued the regression), i can see in previous commits it was splitted by \r\n\r\n or \n\n. 

The email.parser lib already parses the whole header so there is no need for splitting inside the extracted headers. 

Currently the 'content-type:' splitting is causing mails from office365 to be parsed incorrectly by the program, as seen with arc headers..
 arc-message-signature: i=1; a=rsa-sha256; c=relaxed/relaxed;
 d=microsoft.com; s=[REDACTED];
 h=From:Date:Subject:Message-ID:------->Content-Type:<----------MIME-Version:X-MS-Exchange-SenderADCheck; (<- and -> is not actual syntax but to show case it)

Above is 8 lines into the header where as the original header is 224 lines long.

### DIFF
--- a/analyzers/EmlParser/parse.py
+++ b/analyzers/EmlParser/parse.py
@@ -84,12 +84,10 @@ def parseEml(filepath):
     with open(filepath, 'r') as f:
         raw_eml = f.read()
 
-    #parsing the headers with the email library
-    #cause eml_parser does not provide raw headers (as far as I know)
-    #splited string because it was returning the body inside 'Content-Type'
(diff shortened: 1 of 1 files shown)


ch0wm3in:
I've investigated more into this, above is not true the headerparser will sometimes take content with it for unknown reasons.
The better solution would be to add a 'whitespace' to the 'content-type:' index with 'content-type: '

PR #985, opened 2021-05-05
Update fileinfo_analyzer.py

v1p3r0u5:
removed nested self. Caused an error running the analyzer on TheHive 3.5.1.


jeromeleonard:
thank you for your PR. Merging it.

PR #1015, opened 2021-07-01
CIRCL Hashlookup Analyzer

nsmfoo:
CIRCL Hashlookup Analyzer, check your files against a corpus of known good files

https://hashlookup.circl.lu/


adulau:
Any time-frame when this will be merged? We know many users interested into it. Thanks a lot.

PR #1023, opened 2021-08-10
Update Elasticsearch analyzer with unix format 

ajrios33:
It appears that `elk.py` is in DOS format and is causing issues when we import and try to run the analyzer:

`bash: ./elk.py: /usr/bin/env python3^M: bad interpreter: No such file or directory`

We have resolved this issue by doing the following:
`vim elk.py` and `:set ff=unix`

Since we are copying the analyzers from this repo, the analyzer gets reverted on restart.


jiribartusek-ls:
@nadouani is it ready for merge? Sorry for pinging/bothering you, but I am afraid I am new here and still not sure who maintains this... :)

PR #1080, opened 2022-01-26
Case_Reporter responder

cyberpescadito:


azgaviperr:
Tested it and it does the job.

Need maybe some more information on the image size instead of small.

The overall shape of report could be made better as some filed of tables are hard to read (new line) but it does respond to needs of most user that need reporting.

PR #1111, opened 2022-07-19
Virustotal v3

### LINKED ISSUE #1012: [FR] Virustotal Analyzer and VT API v3? (v2 will go offline soon)
The current Virus Total Analyzer and the develop version seem to use the Virus Total API v2. The URLs for this will go offline on October 1st. After that, only the v3 URLs can be used.
Will there be a new analyzer with v3 support?

migueldo:
Update Virustotal analyzer to v3 API and enriched results with new relations.
Resolves #1012


azgaviperr:
Any readon why there is no short report template

Everything else is ok for me

nadouani:
> Any readon why there is no short report template
> 
> 
> 
> Everything else is ok for me 
> 
> 

Because short reports are no longer used :)

migueldo:
> After some extra testing, I can see issues on some characters no showing correctly on the report image

It seems like your fontawesome icons can't load. You can check if they are imported from the developer console of your browser.
I fixed an issue on the date format in the update section of your screenshot but for this one I can't find a way to simulate the `alt` attribute in images.

migueldo:
I changed the icons it should be ok now.

PR #1117, opened 2022-08-02
[FIX] HybridAnalysis API V2

X0x1RG9f:
Fixing HybridAnalysis analyzer that was not working since API V2 :
 - Fix added
 - API Secret not needed anymore in configuration (only API Key)
 - Adding possibility to analyze URL or Domain


bcampru:
Hello @X0x1RG9f I didn't see that there was an oppened pull request for this bug and I also have written a fix, while I was checking the API V2 I've found that there is a field named `threat_score` that in my opinion would make more sense to map with cortex's value, here it is my solution if you want to check it: https://github.com/bcampru/Cortex-Analyzers/blob/master/analyzers/HybridAnalysis/HybridAnalysis_analyzer.py

evan251:
Hello everyone, 
Does anyone know if this fix is planned to be merged to the main branch ? 
Our team would be glad to have this analyzer working again ! 
Best regards,
Evan

PR #1118, opened 2022-08-03
[FIX] No taxonomy was added in case abuse address was found

X0x1RG9f:
Fix added. No taxonomy was added in case abuse address was found.


nusantara-self:
Fixed in #1262

PR #1154, opened 2023-01-27
Update urlcategory.py

lucamemini:
Fix url encoding (get) and check status_code returned by fortiguard webserver


nusantara-self:
Closes #1228

PR #1158, opened 2023-02-10
Added analyzer for Crowdstrike FalconX API

tbi88:
I created an analyzer for the Crowdstrike FalconX API. 

You need to have a paying subscription to use the service.


tbi88:
Could you merge this please?

Issue #1182, opened 2023-04-27
[Bug] OpenCTI Analyser


FormindMPO:
## **Describe the bug**
OpenCTI analyser is not working if any observable found is related to a report

## **To Reproduce**
1. Using any OpenCTI instance, create an observable, with an indicator.
2. Create a report using this indicator (hence this observable) :

![image](https://user-images.githubusercontent.com/127124531/234866413-52e5aa3e-c5ea-43ff-9f9c-cdacc2033ee5.png)

3. Search for this observable using [Cortex-Analyzers/analyzers/OpenCTI](https://github.com/TheHive-Project/Cortex-Analyzers/tree/master/analyzers/OpenCTI)

## **Expected behavior**
Informations about this observable 

## **Error**

```PYTHON
Traceback (most recent call last): 
File "/worker/OpenCTI/opencti.py", line 111, in <module> OpenCTIAnalyzer().run()
File "/worker/OpenCTI/opencti.py", line 97, in run del(report["x_opencti_graph_data"]) 
KeyError: 'x_opencti_graph_data' Killed 
```

## **Complementary information**
/

## **Work environment**
 - Client OS: Windows
- Server OS: Linux
- Browse type and version:
- Cortex version: any
- Cortex Analyzer/Responder name: any
- Cortex Analyzer/Responder version: any
- OpenCTI version : >=5.0.0

## **Possible solutions**
OpenCTI removed the "x_opencti" field 2 yers ago in this [commit d008b33](https://github.com/OpenCTI-Platform/client-python/commit/d008b3346d0f91209470b1b829b6b1bee1c51d54).
You only need to remove line 97 from [opencti.py](https://github.com/TheHive-Project/Cortex-Analyzers/blob/master/analyzers/OpenCTI/opencti.py) : 
```PYTHON
del(report["x_opencti_graph_data"])
```
Then, please release a new dockerhub [opencti_searchexactobservables](https://hub.docker.com/r/cortexneurons/opencti_searchexactobservables) and [opencti_searchobservables](https://hub.docker.com/r/cortexneurons/opencti_searchobservables) version
Thank you !

## **Additional context**
/

PR #1183, opened 2023-04-27
Editing bug related to https://github.com #1182

FormindMPO:
Everything is explained in #1182


ChestoOfGlen:
Surely this is a super simple and easy fix that could be quickly merged and new container images generated by the StrangeBee team?
This is currently a blocker for us running OpenCTI analysers against observables from TheHive as one of our major data sources in OpenCTI includes reports associated with observables.

jeromeleonard (approved):

Issue #1237, opened 2023-12-06
'Triage' analyzer adapation to fit Recorded Future solution (based on Triage)


Tux-Panik:
**Feature description**
In July 2022, Recorded Future acquired Hatching Triage sandbox technology.
Thus, the existing 'Triage' analyzer developed by @nsmfoo could be adapted to fit with Recorded Future's sandbox API and allow submission of files, URLs and IPs as well.

**Describe the solution you'd like**
I've manage this adaptation, but to avoid any copyright infringements, we would prefer not to publish it prior to get validation.
Original Analyzer is under GPLv3, so it shouldn't be an issue to have such light modification while we mention the source.

Tux-Panik:
PR created [](https://github.com/TheHive-Project/Cortex-Analyzers/pull/1252)

PR #1252, opened 2024-03-27
Recorded Future Sandbox Analyzer

Tux-Panik:
In July 2022, Recorded Future acquired Hatching Triage sandbox technology.
Thus, the existing 'Triage' analyzer developed by @nsmfoo could be adapted to fit with Recorded Future's sandbox API and allow submission of files, URLs and IPs as well.

I've manage this adaptation, but to avoid any copyright infringements, I would prefer not to publish it prior to get validation.
Original Analyzer is under GPLv3, so it shouldn't be an issue to have such light modification while we mention the source.

https://github.com/TheHive-Project/Cortex-Analyzers/issues/1237


rpitts-recordedfuture:
Hello @Tux-Panik , thank you for making this suggestion to include the Recorded Future Sandbox. It's an improvement I was also planning to add and in addition to the Recorded Future sandbox API we could also add the public sandbox, https://tria.ge, URL as well.

From a maintainability perspective, I think it would be best to only have one Triage analyzer that is configurable to select between the 3 URL endpoints.

nadouani:
Hi @Tux-Panik, how would you like to manage the comment from @rpitts-recordedfuture ? Thanks for the contribution.

I can see a screenshot showing the report template. Could you please include it in this PR?

rpitts-recordedfuture:
Hi @nadouani and @Tux-Panik , I made the changes I was suggesting to the Triage analyzer to make it easier to maintain 1 analyzer instead of 3. In my [pull request](https://github.com/TheHive-Project/Cortex-Analyzers/pull/1264), the API URL is now a configurable field so that the user can set it up for Recorded Future Sandbox usage, private sandbox usage, and now free users can configure the public sandbox to make submissions.

Let me know how you would like to proceed. Thank you for considering my changes and suggestions.

Tux-Panik:
Hi man,

Thanks for following-up on this topic and for implementing the changes we
both wanted.

I'll take some time to test it on my own, especially if the code works in
an environnement where it was not crafted.

Will give you my feedback and will let Nabil move forward.
A topic that we could have been discussed face to face past week 😁

See ya
Julien

nusantara-self:
Hello,

Merged #1264 from @rpitts-recordedfuture , which I believe fixes main issues encountered while maintaining a single analyzer

PR #1253, opened 2024-04-10
Update opencti.py

Noatun:
We installed OpenCTI and realized that the Analyzer is no longer working.  Failing with error: got invalid value [{ key: \"objectContains\", values: [Array] }]; Field \"mode\" of required type \"FilterMode!\" was not provided. I looked at the examples here: https://github.com/OpenCTI-Platform/client-python/blob/master/examples/get_reports_about_intrusion_set.py and tried to repair the analyzer.


nadouani:
Hi @Noatun Thanks for the PR 🙏🏼

nadouani:
Hi @Noatun Just accepted the PR #1245

PR #1262, opened 2024-06-04
Quick updates

vpiserchia:


analyzers/Abuse_Finder/requirements.txt
```diff
@@ -1,3 +1,4 @@
 cortexutils
+git+https://github.com/monoidic/ipwhois.git
```
nadouani, line 2: @vpiserchia What's the need for this dependency?

nadouani:
Hi @vpiserchia I just rebased the PR on `develop` branch and left some review comments

analyzers/Abuse_Finder/requirements.txt
vpiserchia: By using this forked version of the library a newer version of the `dnspython` (v2+) lib is installed. This way all analyzers dependencies are harmonized and can be installed in the same docker or system.

analyzers/MISPWarningLists/warninglists_create_db.py
```diff
@@ -20,7 +20,7 @@
 conn_string = "<insert_postgres_conn_strin>"
 warninglists_path = "misp-warninglists/**/list.json"
 
-engine = create_engine(conn_string, use_batch_mode=True)
+engine = create_engine(conn_string))
```
vpiserchia, line 23: there is a syntax error. need to fix

analyzers/Abuse_Finder/requirements.txt
nusantara-self: Hello! I'm removing it from requirements as package abuse_finder already installs dnspython v2+ (specifically, v2.7.0). It also installs the official ipwhois v1.3.0, which is current latest version; no issues observed with the analyzer on my end.

Issue #1277, opened 2024-09-03
New Analyzer - LupovisProwlAnalyzer


stacsirt:
**Feature description**

The LupovisProwlAnalyzer is an analyzer for TheHive that integrates with the Lupovis Prowl API to evaluate the reputation of IP addresses

Features:
    IP Reputation Analysis: Checks if an IP address is linked to malicious activities.
    Detailed Reporting: Generates both detailed and summary reports for analysis results.
    Flexible Configuration: Supports configuration of API keys and proxy settings.

**Describe the solution you'd like**
Once installed and configured you can action any IP address in TheHive with this analyzer and this will interact with the Lupovis Prowl API key to evaluate the IP address and tell if the IP is malicious or not and if it is malicious it will tell you what it is known to be linked to.

Below is a link to the config for the analyzer:

https://github.com/stacsirt/ProwlAPI---IP-reputation-check

PR #1278, opened 2024-09-04
Added new LupovisProwlAnalyzer #1277

stacsirt:
Refer to the issue https://github.com/TheHive-Project/Cortex-Analyzers/issues/1277


nusantara-self:
Hello @stacsirt,

Thanks for the PR. Have you had a test environment to try and test out the analyzer? I've made multiple changes to at least make it run-able, there were a lot of issues including syntax. 

However, i'm a bit confused about the run function. There is, I think, a confusion with taxonomies (this needs to be managed in the `build_taxonomy` function, not in the `run` function) but also in the main code logic in `run`.

The code as it is now, even with invalid key, will return a "valid report", whereas it should use `self.error("some string ...")` in those cases. I am also not sure why making it complicated to use `subprocess` instead of just using `requests` module.

If there is any difficulty regarding that part, I may help if I have information on what are some resulting payloads from LupovisProwl.

Thanks in advance,
Regards

analyzers/LupovisProwl/LupovisProwlAnalyzer.py
```diff
@@ -0,0 +1,94 @@
...
+                output = stdout.decode()
+                # Parse the JSON output
+                output_json = json.loads(output)
+                # Check if "ttps" field is empty or contains "no known malicious activity"
+                details = output_json.get('ttps', 'No known malicious activity')
+                level = 'malicious' if output_json.get('ttps') else 'safe'
+                result = {
+                    "ip": ip,
+                    "details": details,
+                    "taxonomy": self.build_taxonomy(level, 'LupovisProwl', 'Threat', details)
```
nusantara-self, line 85: This is not needed. Taxonomies (resulting in short reports in TheHive) are solely managed by the build_taxonomy function

analyzers/LupovisProwl/LupovisProwlAnalyzer.py
```diff
@@ -0,0 +1,94 @@
...
+        try:
+            # Constructing the curl command
+            curl_command = ['curl', '-H', f'x-api-key: {self.prowl_api_key}', url]
+            # Executing the curl command and capturing output
+            process = subprocess.Popen(curl_command, stdout=subprocess.PIPE, stderr=subprocess.PIPE, env={'http_proxy': self.proxy, 'https_proxy': self.proxy})
+            stdout, stderr = process.communicate()
+            if process.returncode == 0:
+                output = stdout.decode()
+                # Parse the JSON output
+                output_json = json.loads(output)
```
nusantara-self, line 78: Please use `requests` to make the call & use status code to either pursue and self.report your results (just like below) or self.error("my error message with details"), for exemple in case of 403 Forbidden.

analyzers/LupovisProwl/LupovisProwlAnalyzer.py
stacsirt: Hi @nusantara-self,

Thanks for your overview of the script, apologies for the syntax issues, I have now made the amendments to align with the cortex analyzer guidelines.

Thanks again,
Lyle

nusantara-self: Lovely, thanks a lot Lyle! Going to test it soon and release! :)

stacsirt: I tested it in our environment and it worked perfectly :)

PR #1279, opened 2024-09-05
Adapt filters to OpenCTI v6

jmbodelon:
https://docs.opencti.io/latest/reference/filters-migration/


nusantara-self:
Hello @jmbodelon,

Thank you for your PR. FYI, PR #1245 has been previously merged, which should address the same issue you highlighted. This fix is now available in version 3.3.8 of Cortex-Analyzers.

Please let us know if the analyzer now works as expected, or if you think further updates might be necessary.

Regards,

Issue #1290, opened 2024-10-25
[Bug] Elasticsearch analyzer does not work with index that has no @timestamp field


gaistou:
**Describe the bug**

Elasticsearch analyzer does not work with Elastic indices that do not have a @timestamp field.

**To Reproduce**

- create an Elastic index without @timestamp field
- add the index in the list of indices searched by the analyzer
- run the analyzer
- you get no result and an error about sorting on @timestamp

**Expected behavior**

Elasticsearch analyzer should be able to match on indices that have no @timestamp fields.

**Complementary information**

The analyzer tries to sort by @timestamp. When there is no@timestamp field this sort fails and generates an error.

**Work environment**
 - Cortex Analyzer/Responder name: ElasticSearch
- Cortex Analyzer/Responder version:1.0

vpiserchia:
@gaistou  I just pushed a change that should solve the issue you're encountering.

I have also to say that I do not see why you have this use case. This Analyzer accomplish for historical/retro searches; which do not really make sense without a timestamp to use...
let me know

vito

PR #1322, opened 2025-02-06
Update Elasticsearch Analyzer to not fail when @timestamp has no mapping in searched indices

vpiserchia:
As described in: #1290


nusantara-self:
Closes #1290

Issue #1326, opened 2025-02-14
[ENG] Feature Request: LdapQuery could import Observables, Tags, Custom Fields


kiaora17:
**Feature description**
The LdapQuery analyzer allows to collect valuable information about the users of the directory. Among this information, some can be used to facilitate:

- the transition from a username to an email address by automatically importing the corresponding Observables,
- the filtering of observables by using automatically imported tags (e.g. which users are still `Active`),
- the construction of reliable dashboards (e.g. by automatically populating a custom field `country`).

**Describe the solution you'd like**
To do this, the analyzer could offer the possibility to:

- define which attributes could be imported as **Observables**,
- define which attributes could be imported as **Tags**,
- define which attributes could be imported as **Custom Fields**.

Additionally, to avoid flooding the LDAP server when processing large lists of usernames or email addresses, some filters could be implemented to only send relevant queries to the LDAP server.

**Additional context**
These feature requests would notably help to handle Credential Leak cases more efficiently.

PR #1327, opened 2025-02-14
LdapQuery3.0: Import Observables, Tags and Custom Fields

kiaora17:
Refer to the issue [#1326](https://github.com/TheHive-Project/Cortex-Analyzers/issues/1326)


kiaora17:
Hello @nusantara-self,

By [deleting the Dockerfile](https://github.com/kiaora17/Cortex-Analyzers/commit/1c0c673bf1df63a654257f09fff493931789a90c#diff-e651765a91a6c319914b6d3b0ffea08053c3631dbd11aeffd412cd50bdab70e3L15) that was present in initial PR, it seems that the docker image of _LdapQuery3.0_ does not support TLS connection to reach LDAP server, as you can see in error message example shared below.

In fact, it seems to use SSLv3 instead, which is deprecated for a long time now. This is why the specific Dockerfile provided in the PR was installing **openssl**.

Can you please check this behavior on your side and update the Dockerfile for _LadapQuery3.0_, or provide a way to use analyzer's docker image with TLS encryption?

```
{
  "errorMessage": "Error during LDAP connection: (\"('socket ssl wrapping error: [SSL: SSLV3_ALERT_HANDSHAKE_FAILURE] ssl/tls alert handshake failure (_ssl.c:1028)',)\",)",
  "success": false
}
```

nusantara-self:
Hello @kiaora17,

Thanks a lot for reporting, this is really useful and an oversight; fixing it asap!

Best,

kiaora17:
Hello @nusantara-self,

As Dockerfile is still missing, we still have the following error when using actual Docker image:
`Error during LDAP connection: ("('socket ssl wrapping error: [SSL: SSLV3_ALERT_HANDSHAKE_FAILURE] ssl/tls alert handshake failure (_ssl.c:1010)',)",)`

In fact, it seems to use SSLv3 instead, which is deprecated for a long time now. This is why the specific Dockerfile provided in the PR was installing openssl.

Can you please put back the Dockerfile in Analyzer's folder.

Thank you.

Issue #1155, opened 2023-01-27
Umbrella Analyzer API not supported


Realgagenichols:
Umbrella Analyzer API not supported. Use this template for bug reports. If you'd like to request a feature, please be as descriptive as possible and delete the template except the first section (Request Type)

### Request Type
(select Bug, Analyzer or Feature and **remove this line**)
Analyzer

### Work Environment
N/A

### Description
The Umbrella analyzer API is no longer supported and has been deprecated for V2 

### Possible Solutions
Update the Umbrella analyzer to work with V2 API, update base URI, V2 uses a bearer token generated from the API key and secret now, new paths and functionality, etc

### Complementary information
(add anything that can help identifying the problem such as **logs**, **screenshots**, **configuration dumps** etc.)

PR #1330, opened 2025-02-27
Update Umbrella.py

Noatun:
This script has initialy been written for the Umbrella API v1. This API has been retired since March 22.
We recently had the need to use it, so I changed the code to work with the v2 API.


nusantara-self:
Hello @Noatun,

Great job! 

Fixes #1155

Issue #1138, opened 2022-11-14
[Bug] Analyzer Threatcrowd_1_0 not working


dominiksr:
**Describe the bug**
Analyzer Threatcrowd_1_0  not working. There is faulty execution and explanatory error.

**To Reproduce**
1. New Analysys
2. Choose data type IP or mail or domain
3. Enter data eg 1.1.1.1

**Expected behavior**
Result of the analysis.

**Complementary information**
err:
Unexpected Error: HTTPSConnectionPool(host='www.threatcrowd.org', port=443): Max retries exceeded with url: /searchApi/v2/ip/report/?ip=1.1.1.1 (Caused by SSLError(SSLCertVerificationError("hostname 'www.threatcrowd.org' doesn't match '*.otxb.io'")))

**Work environment**
- Server OS: Ubuntu
- Cortex version: 3.1.7-1 deb
- Cortex Analyzer name: Threatcrowd_1_0  

**Possible solutions**
Probably renewing the certificate would make it work again.

**Additional context**
The analyzer stopped working about 2 months ago.

jeromeleonard:
the service is not available anymore at the address www.threatcrowd.org and the certificate is invalid.

dominiksr:
Will this analyzer be removed from the cortex analyzer database?

nusantara-self:
Hello,

Analyzer has been removed from catalog from now on. #133

PR #1334, opened 2025-03-10
Deprecation support for analyzers & responders in CI

nusantara-self:
Adding `.deprecated` to **_no-longer supported // working analyzer or responder_** `json` file excludes it from catalog & build.


nusantara-self:
Closes #1138

PR #1350, opened 2025-04-17
Fix analyzer: incompatible ctx.io api [Malwares]

Ignatella:
Hi!

`https://www.ctx.io/about/api` is now incompatible with the Malwares analyzer. This PR aims to fix it


nusantara-self:
Hello @Ignatella,

Thanks for the PR! I've noticed the domain name `https://api.ctMalwares_Scan.json.io/v1/` does not match the documentation`https://api.ctx.io/v1/`. Could you please confirm if this is a typo?

Thanks a lot,
Fabien

Ignatella:
Yes, this really looks like my fault, idk what this url comes from. I will change and verify it. Thank you for the review :)

Ignatella:
@nusantara-self Hello, just changed the api url

PR #1408, opened 2026-01-09
Better context for analyzers & responders - integration type / subscription info

nusantara-self:
Integration type : 
- local
- external_api
- self_hosted
- hybrid


nusantara-self:
batch 1

PR #1416, opened 2026-01-22
Update MS Defender Responders

vpiserchia:
- Simplify submit indicator code following the latest API definitions.
- added missing modes (Warn, BlockAndRemediate, Audit, Allowed)
- deprecate the Alert only mode


nusantara-self:
Great PR, thanks for the contribution!

Issue #1291, opened 2024-10-25
[Bug] Elasticsearch "keys" field is required but should not


gaistou:
**Describe the bug**

The description of the analyzer says that the "keys" param is not required. However, leaving it empty when running the analyzer with a local config files generates an error.

**To Reproduce**

- create a local input.json file without a "keys" field
- run the analyzer with the local input.json
- you get a "cannot iterate on None type" error

**Expected behavior**

The missing "keys" param should not generate an error.

**Complementary information**

The issue comes from the "zip" function call on index,keys,users, and passwords.

**Work environment**

 - Cortex Analyzer/Responder name: ElasticSearch
- Cortex Analyzer/Responder version: 1.0

PR #1455, opened 2026-06-09
Fix Elasticsearch analyzer TypeError when optional configurations are…

jeffrey-theog06:
This PR fixes issue #1291, where the Elasticsearch analyzer raises a TypeError when optional configuration parameters (keys, users, or passwords) are omitted.

The issue occurred because Python's built-in zip() function was called on self.endpoints, self.keys, self.users, and self.passwords without checking if the optional fields were configured. If any of these fields were left empty (resolving to None), the zip() call raised a TypeError.

Key Changes:
Imported zip_longest from the standard itertools library.
Normalized Parameter Lists: Handled cases where optional configurations are None or string values, converting them to lists.
Graceful Iteration: Replaced zip with zip_longest(..., fillvalue=None) to loop through configured endpoints and their credentials, even if their list lengths are mismatched or unconfigured.
Testing Done
Checked syntax compilation of elk.py (no errors).
Validated normalization logic under multiple configurations (default unconfigured optionals, string endpoint with None optionals, mismatched credential list lengths).
Verified the analyzer's execution path using mock endpoints and mock credentials without raising a TypeError.

### DIFF
--- a/analyzers/Elasticsearch/elk.py
+++ b/analyzers/Elasticsearch/elk.py
@@ -5,6 +5,7 @@ from cortexutils.analyzer import Analyzer
 import dateutil.parser
 from datetime import datetime
 from packaging.version import Version
+from itertools import zip_longest
 
 # utils
 import operator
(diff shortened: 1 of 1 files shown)


jeffrey-theog06:
Hi @nusantara-self Please check on My PR and let me know any changes need to be done or its good to go for merge.. Since i havent heared any news from this..

